PR #2, opened 2022-04-22
fix(action): Fix Docker caching permission error

Kurt-von-Laven:
GitHub Actions doesn't grant us permission to cache the Docker root directory even if Docker is running in rootless mode. Instead, on cache miss, export the images to a TAR archive at a path we have permission to cache and subsequently load from on cache hit. Composite actions don't yet support post steps, so port to a TypeScript action since Docker images can't be saved until after the calling workflow pulls or builds them. Use Yarn Zero-Installs, because dependencies of GitHub Actions must be version controlled, and Yarn Zero-Installs works by compressing and tracking all Node.js dependencies. Check in the `dist` directory output by TypeScript since GitHub Actions requires its presence in order to run the action and cannot directly execute TypeScript. Add `cache-hit` output so callers can branch if needed.


mwarres (approved):
Looks good as far as I can tell

PR #3, opened 2022-04-23
fix(action): Port from Yarn Zero-Installs to ncc.

Kurt-von-Laven:
GitHub Actions doesn't support the former, so the action crashes when used, because the dependencies aren't found. ncc compiles all dependencies into a single index.js file. Do this for both the main and post steps. Since `ncc` doesn't allow the output filename to be specified, only the output directory, put each `index.js` in a different subdirectory of `dist`. Continue using Yarn, but remove its cache to improve shallow cloning performance when this action is used.

Inform when post step has nothing to do. Most of the official GitHub cache action's logs wind up in our logs by virtue of using their API. However, we make no cache API call in the post step on cache hit. Manually mirror the cache action's log output in this case.

Use a shell available in GitHub Actions in `util.ts`. `/usr/bin/env bash` is not found on the `ubuntu-20.04` GitHub-hosted runner. Use `/usr/bin/bash` instead, which is.


mwarres (approved):
Looks good as far as I can tell!

PR #24, opened 2022-05-19
ci(workflow): Check out full repository history

Kurt-von-Laven:
In the Test workflow, the pre-commit-action now runs the Commitizen action, which relies on Git tags. The official GitHub checkout action does a sparse checkout by default, which does not include tags, causing version bumps to fail. Pass `fetch-depth: 0` to instruct the checkout action to perform a full checkout, including tags.


mwarres (approved):
LGTM!

PR #28, opened 2022-05-20
chore(GitHub-Actions): Upgrade to latest versions

Kurt-von-Laven:
GitHub Actions recently released support for Ubuntu 22.04 and sharing caches across OS versions can cause build failures. Pull in a fix to pre-commit-action to prevent such sharing. Use the same new cache key as pre-commit-action to improve CI performance now that this won't cause build failures. This spares saving the cache twice on miss and invalidates all caches both presently and on OS upgrades.

ScribeMD/pre-commit-action 0.7.0 --> 0.7.1

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,25 +9,24 @@ on:
 jobs:
   test:
     name: Test
-    strategy:
-      matrix:
-        runner:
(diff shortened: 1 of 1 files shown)


mwarres (approved):
LGTM!

PR #29, opened 2022-05-20
fix(post): Save cache on cache miss

Kurt-von-Laven:
`getState` always returns a string, so the boolean values we pass to `saveState` are coerced to `"false"` and `"true"`. Both of these are truthy values, so the action always incorrectly stated that a cache hit occurred without ever saving the cache. Test explicitly for the string `"true"` rather than for truthiness so that the cache is saved on miss.

When saving Docker images on cache miss, redirect `xargs --verbose` output from stderr to stdout since it typically isn't an error.


mwarres (commented):
Looks like this isn't working properly, see this error in the post cache docker image section of the log:
Error: docker save --output /home/runner/.docker-images.tar megalinter/megalinter-javascript:v5.13.0

Kurt-von-Laven:
Good catch. Fixed!

mwarres (approved):
LGTM!

PR #36, opened 2022-05-24
chore(Yarn): Upgrade all dependencies to latest

Kurt-von-Laven:
Upgrade all transitive dependencies as well.

@actions/cache ^2.0.2 --> ^2.0.4
@actions/core ^1.6.0 --> ^1.8.2
@types/eslint ^8.4.1 --> ^8.4.2
@types/node ^17.0.25 --> ^17.0.35
@types/prettier 2.6.0 --> 2.6.1
@yarnpkg/sdks ^2.5.0 --> ^2.6.2
eslint ~8.13.0 --> ~8.16.0
typescript ^4.6.3 --> ^4.6.4


mwarres (commented):
There seemed to me to be upgrades to dependencies that weren't mentioned in the commit message.  Not sure if this is problematic or a transitive dependency situation???

Kurt-von-Laven:
Yeah, these are transitive dependencies. The purpose of a lock file is to pin the exact version of not only direct dependencies, but transitive dependencies as well. This offers a high degree of reproducibility. I did forget to mention that the transitive dependencies were updated in the commit message though, so I added that.

mwarres (approved):
LGTM

PR #37, opened 2022-05-29
refactor(action): Separate scripts from modules

Kurt-von-Laven:
Move implementations of `loadDockerImages` and `saveDockerImages` from `main.ts` and `post.ts` to a new `docker.ts` file. Leave only the top-level awaits in `main.ts` and `post.ts` so that `docker.ts` can be imported without side-effects. This is a requirement for using `testdouble.replaceEsm`, which we intend to use. Specify extensions of local filename imports for the same reason. Use "node" scope to explicitly import built-in modules, most notably to disambiguate the built-in util from the local util module.


davisgtaylor (approved):
no issues noted

PR #38, opened 2022-06-04
chore(pre-commit): Upgrade all hooks to latest

Kurt-von-Laven:
ScribeMD/pre-commit-hooks 0.5.0 --> 0.5.1

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
 
   ## Python, Polyglot, Git, pre-commit
   - repo: https://github.com/ScribeMD/pre-commit-hooks
-    rev: 0.5.0
+    rev: 0.5.1
     hooks:
       - id: no-merge-commits
       - id: asdf-install


davisgtaylor (approved):

davisgtaylor:
The test failing here is expected yes?

Kurt-von-Laven:
Actually no, so thank you for pointing that out. I re-ran the test suite, and it passed this time. Based on the original logs, the error most likely occurred in the `saveCache` method of `@actions/cache`, but the error itself wasn't logged. I'm guessing it was a network failure while uploading the cache, and in any case it's entirely unrelated to the change in this pull request. If we manage to reproduce this issue with [debug logging](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging) enabled, I will file a bug report on [the GitHub cache action](https://github.com/actions/cache). Otherwise, I don't think they'll be able to investigate this anymore than I can.

PR #49, opened 2022-06-14
chore: Add and configure Dependabot

mwarres:
Automatically update dependencies via Dependabot
Note: automatic updates exclude pre-commit, asdf, and Yarn v2+ and its dependencies; these are not supported by Dependabot.


Kurt-von-Laven (approved):
**LGTM**

PR #56, opened 2022-06-18
test(Jest): Enforce 100% test coverage

Kurt-von-Laven:
Enforce 100% test coverage on all files matching `src/**/*.ts` in CI. When running locally, only collect test coverage from files related to the specific tests that were run to avoid false positives. When running as a pre-commit hook with `--findRelatedTests`, Jest automatically limits code coverage collection to the modified files. Add a 100% test coverage badge to the README with the 24 pixel base64 trophy Octicon SVG icon embedded. Match the label color of the Jest badge (`#FF8C42`).


davisgtaylor (approved):

mwarres (approved):
LGTM!

PR #60, opened 2022-06-20
style(docker.test): Specify a return type

Kurt-von-Laven:
This allows TypeScript to enforce that a number is returned from the comparison function.


mwarres (approved):
LGTM!

PR #63, opened 2022-06-22
docs(README): Use Coolors' suggestions for badges

Kurt-von-Laven:
Configure [Coolors](https://coolors.co/) with the logo color as the primary color, label background color (`#555555`) as the secondary color, and text color (`#FFFFFF`) as the quaternary color. Use the first recommended tertiary color for the message background color.


mwarres (approved):
Looks good overall! Might look slightly nicer if the Test and pre-commit badges weren't adjacent because the Test badge is green when passing and the pre-commit badge is also green, but a slightly different shade.

PR #79, opened 2022-07-10
docs(README): Document required permissions

Kurt-von-Laven:
Configure markdown-link-check to accept any compression to avoid false positive on GitHub Docs link.


mwarres (approved):
LGTM

PR #81, opened 2022-07-13
feat(action): Add `read-only` input

### LINKED ISSUE #44: Stage "Post Cache Docker images" fails for all runners except one when missing cache
I added the step from the Usage section of `README.md` before the first use of Docker (https://github.com/contiki-ng/contiki-ng/pull/1946). The second run that hits the cache works, but all runners except one fails the first run when they don't get a cache-hit. The runners fail in the "Post Cache Docker images" stage with a message similar to:

```
Post job cleanup.
docker image list --format "{{ .Repository }}:{{ .Tag }}" | 2>&1 xargs --delimiter="
" --no-run-if-empty --verbose --exit docker save --output ~/.docker-images.tar
docker save --output /home/runner/.docker-images.tar contiker/contiki-ng:63f7e91e3 node:14 node:16 buildpack-deps:stretch buildpack-deps:buster buildpack-deps:bullseye debian:9 debian:10 debian:11 node:14-alpine moby/buildkit:latest ubuntu:20.04 ubuntu:18.04 node:16-alpine node:12 node:12-alpine alpine:3.12 alpine:3.13 alpine:3.14 ubuntu:16.04

/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/contiki-ng/contiki-ng --files-from manifest.txt
file:///home/runner/work/_actions/ScribeMD/docker-cache/0.1.4/dist/post/index.js:1
eateRequire as e}from"module";var t={1312:(e,t,a)=>{Object.defineProperty(t,"__esModule",{value:true});function _interopDefault(e){return e&&typeof e==="object"&&"default"in e?e["default"]:e}var i=_interopDefault(a(2781));var n=_interopDefault(a(3685));var o=_interopDefault(a(7310));var r=_interopDefault(a(3621));var s=_interopDefault(a(5687));var p=_interopDefault(a(9796));const l=i.Readable;const d=Symbol("buffer");const c=Symbol("type");class Blob{constructor(){this[c]="";const e=arguments[0];const t=arguments[1];const a=[];let i=0;if(e){const t=e;const n=Number(t.length);for(let e=0;e<n;e++){const n=t[e];let o;if(n instanceof Buffer){o=n}else if(ArrayBuffer.isView(n)){o=Buffer.from(n.buffer,n.byteOffset,n.byteLength)}else if(n instanceof ArrayBuffer){o=Buffer.from(n)}else if(n instanceof Blob){o=n[d]}else{o=Buffer.from(typeof n==="string"?n:String(n))}i+=o.length;a.push(o)}}this[d]=Buffer.concat(a);let n=t&&t.type!==undefined&&String(t.type).toLowerCase();if(n&&!/[^\u0020-\u007E]/.test(n)){this[c]=n}}get size(){return this[d].length}get type(){return this[c]}text(){return Promise.resolve(this[d].toString())}arrayBuffer(){const e=this[d];const ...
```

mwarres:
Allow `read-only` option for cache. Use `read-only` option to eliminate warning upon attempted duplicate cache save in Test workflow. The `read-only` option also allows parallel job runs without spurious warnings from multiple attempted cache saves.

Fixes #44.


.github/workflows/test.yaml
```diff
@@ -28,6 +28,7 @@ jobs:
             docker-${{ steps.os.outputs.image }}-${{
               hashFiles('.pre-commit-config.yaml')
             }}
+          read-only: "true"
```
Kurt-von-Laven, line 31: I'm not sure, but I think the quotes may be unnecessary?

.github/workflows/test.yaml
```diff
@@ -38,10 +39,69 @@ jobs:
...
+  restore-cache:
+    name: Restore Cache
+    needs:
+      - save-cache
+    runs-on: ubuntu-20.04
+    permissions: {}
+    steps:
+      - name: Check out repository.
+        uses: actions/checkout@v3.0.2
+      - name: Load test cache.
```
Kurt-von-Laven, line 65: nit: Consider "Load Docker images" for parity with the corresponding `save-cache` step.

.github/workflows/test.yaml
```diff
@@ -38,10 +39,69 @@ jobs:
...
-      - name: Send Slack notification with job status.
-        if: always()
+  save-cache:
+    name: Save Cache
+    runs-on: ubuntu-20.04
+    permissions: {}
+    steps:
+      - name: Check out repository.
+        uses: actions/checkout@v3.0.2
+      - name: Cache Docker images.
```
Kurt-von-Laven, line 49: nit: Consider "Save Docker images" to specify that this step is never intended to load Docker images.

.github/workflows/test.yaml
```diff
@@ -38,10 +39,69 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    permissions: {}
+    steps:
+      - name: Send Slack notification with workflow result.
         uses: ScribeMD/slack-templates@0.6.2
         with:
-          bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
-          channel-id: ${{ secrets.SLACK_ACTIONS_CHANNEL_ID }}
+          bot-token: ${{ secrets.SLACK_TEMPLATES_BOT_TOKEN }}
+          channel-id: ${{ secrets.SLACK_TEMPLATES_CHANNEL_ID }}
```
Kurt-von-Laven, line 104: Please revert since these secret names come from the slack-templates documentation and are for explanatory purposes only.

README.md
```diff
@@ -65,6 +67,13 @@ The explicit cache key to ferry to the
 `restore-keys` are not supported, because partial cache restoration leads to a
 ["snowball" effect](https://glebbahmutov.com/blog/do-not-let-npm-cache-snowball/).
 
+### Optional
+
+#### `read-only`
+
+Use this option to enable loading the docker-cache upon cache hit and disable
+saving the docker-cache upon cache miss.
```
Kurt-von-Laven, line 75: Multiple instances: this can be read to imply that this input has a bearing on the load step. It's generally advisable in the case of Boolean arguments to indicate which setting has which behavior and what the default is. See [the documentation for pre-commit-action's `bump` input](https://github.com/ScribeMD/pre-commit-action#bump) for an example.

action.yaml
```diff
@@ -11,6 +11,11 @@ inputs:
       are not supported, because partial cache restoration leads to a "snowball"
       effect.
     required: true
+  read-only:
+    description: >
+      An optional parameter which enables loading the docker-cache upon cache hit and disables saving the docker-cache upon cache miss
```
Kurt-von-Laven, line 16: `required: false` below should suffice to indicate that this input is optional.

src/docker.test.ts
```diff
@@ -11,15 +11,15 @@ jest.unstable_mockModule("./util.js", (): typeof import("./util.js") => ({
...
+  const mockCallCounts: Record<string, number> = {};
   const callOrders = mocks.map((currentMock: jest.Mock): number => {
-    expect(currentMock).toHaveBeenCalledTimes(1);
-    return <number>currentMock.mock.invocationCallOrder[0];
+    const mockName = currentMock.getMockName();
+    const callCount = mockCallCounts[mockName] ?? 0;
+    mockCallCounts[mockName] = callCount + 1;
+    return <number>currentMock.mock.invocationCallOrder[callCount];
   });
-
```
Kurt-von-Laven, line 22: Nit: undo whitespace remove for readability.

src/docker.test.ts
```diff
@@ -47,14 +47,18 @@ describe("Docker images", (): void => {
     cache.restoreCache.mockResolvedValueOnce(cacheHit ? KEY : undefined);
     await docker.loadDockerImages();
 
-    expect(core.getInput).lastCalledWith("key", { required: true });
+    expect(core.getInput).lastCalledWith("key", { required: true }); // TODO: or read-only
     expect(cache.restoreCache).lastCalledWith([util.DOCKER_IMAGES_PATH], KEY);
   };
 
-  const mockedSaveDockerImages = async (cacheHit: boolean): Promise<void> => {
+  //helper function to save docker images
```
Kurt-von-Laven, line 54: This is a fairly redundant comment. Suggest either expanding or simply dropping it. Also, remember the space and capitalization after `//`.

src/docker.test.ts
```diff
@@ -47,14 +47,18 @@ describe("Docker images", (): void => {
...
+  const mockedSaveDockerImages = async (
+    cacheHit: boolean,
+    readOnly: boolean = false
+  ): Promise<void> => {
     core.getState.mockReturnValueOnce(cacheHit.toString());
+    core.getInput.mockReturnValueOnce(readOnly.toString());
     await docker.saveDockerImages();
-
```
Kurt-von-Laven, line 57: Nit: undo whitespace remove for readability.

src/docker.test.ts
```diff
@@ -110,13 +114,33 @@ describe("Docker images", (): void => {
...
+    await mockedSaveDockerImages(true, true);
+
+    expect(core.info).lastCalledWith(
+      `Cache hit occurred on the primary key ${KEY}, not saving cache.`
+    );
+    expect(util.execBashCommand).not.toHaveBeenCalled();
+    expect(cache.saveCache).not.toHaveBeenCalled();
+  });
+
+  test("aren't saved on cache hit when read-only false", async (): Promise<void> => {
```
Kurt-von-Laven, line 127: Nit: "false" --> "is false" for consistency.

src/docker.test.ts
```diff
@@ -110,13 +114,33 @@ describe("Docker images", (): void => {
...
+
+  test("aren't saved on cache miss with read-only true", async (): Promise<void> => {
+    await mockedSaveDockerImages(false, true);
+
+    expect(core.info).lastCalledWith(
+      `Cache miss occurred on primary key ${KEY}.  Not saving cache as read-only option selected.`
+    );
+    expect(util.execBashCommand).not.toHaveBeenCalled();
+    expect(cache.saveCache).not.toHaveBeenCalled();
+  });
```
Kurt-von-Laven, line 145: This feels like it belongs below the other cache miss case.

src/docker.ts
```diff
@@ -15,9 +15,14 @@ const loadDockerImages = async (): Promise<void> => {
 };
 
 const saveDockerImages = async (): Promise<void> => {
+  const readOnly = getInput("read-only");
   const key = getInput("key", { required: true });
   if (getState(CACHE_HIT) === "true") {
     info(`Cache hit occurred on the primary key ${key}, not saving cache.`);
+  } else if (readOnly === "true") {
+    info(
+      `Cache miss occurred on primary key ${key}.  Not saving cache as read-only option selected.`
```
Kurt-von-Laven, line 24: "on" --> "on the"

Kurt-von-Laven: Nit: prefer one space after periods in digital copy for readability.

Kurt-von-Laven, line 24: Nit: "selected" --> "was selected"

Kurt-von-Laven (changes requested):
Off to a great start! Don't forget to link the pertinent GitHub issue and squash the last two commits into the first one.

.github/workflows/test.yaml
```diff
@@ -38,10 +39,68 @@ jobs:
...
+    steps:
+      - name: Check out repository.
+        uses: actions/checkout@v3.0.2
+      - name: Save Docker images.
+        uses: ./
+        with:
+          key: docker-cache-test
+      - name: Build an empty test Docker image.
+        run: |
+          docker build --tag empty .
```
Kurt-von-Laven, line 55: I didn't realize that building any image whatsoever, even one from scratch creates a lot of images, and the resulting cache size is ~1,373 MB. Let's just pull a small image (maybe an alpine image?) down from GHCR (GitHub Container Registry) instead.

Kurt-von-Laven (changes requested):
I'm afraid I don't understand what's causing the permission error on cache deletion.

.github/workflows/test.yaml
```diff
@@ -38,10 +39,68 @@ jobs:
...
+    needs:
+      - save-cache
+      - restore-cache
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: write # for deleting cache
+    steps:
+      - name: Delete existing test cache.
+        run: >
+          curl
```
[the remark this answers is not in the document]
mwarres: Will do.  I looked at the docs referred to on the failure and linked to [https://docs.github.com/en/rest/actions/permissions](url).  Under [Set GitHub Actions permissions for an enterprise](https://docs.github.com/en/rest/actions/permissions#set-github-actions-permissions-for-an-enterprise), there is a section that states that the parameter `enabled_organizations` is required.  I'm not seeing that in our code.  They have sample code:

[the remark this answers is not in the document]
mwarres: Cool, so looks like I should just add the following arguments to the curl:
--fail
--silent
--show-error
I'll do that--let me know if that's incorrect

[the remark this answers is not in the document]
mwarres: Is there a convention to the order for the curl arguments?  My googling was not revealing on that front.  I tentatively put the --fail --silent --show-error after the request and headers but before the url

.github/workflows/test.yaml
```diff
@@ -38,10 +39,68 @@ jobs:
...
+          --request DELETE
+          --header "Accept: application/vnd.github.v3+json"
+          --header "Authorization: token $GITHUB_TOKEN"
+          "https://api.github.com/repos/ScribeMD/docker-cache/actions/caches?key=docker-cache-test&ref=$GITHUB_REF"
+  notify:
```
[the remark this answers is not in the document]
mwarres: Do you mean to make it part of the delete-cache job?  I recall that we had separated out the notify job since it was doing the notifications for all of the above jobs.  What am I missing??

action.yaml
```diff
@@ -11,6 +11,11 @@ inputs:
       are not supported, because partial cache restoration leads to a "snowball"
       effect.
     required: true
+  read-only:
+    description: >
```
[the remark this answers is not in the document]
mwarres: Good call.  I noticed when fixing this that the current description is not in the imperative. Is the convention for descriptions to use the imperative or is that more so for other forms of documentation.  Happy to change it if it's appropriate.

.github/workflows/test.yaml
Kurt-von-Laven: [Zipkin is a distributed tracing system](https://zipkin.io/), and they don't look to be in the business of maintaining Docker images, so there's a risk that they might delete that image that may be intended for internal consumption only. [LinuxServer's base images](https://github.com/linuxserver?q=baseimage) look more promising as [LinuxServer maintains a fleet of Docker images](https://www.linuxserver.io/). The smallest of their images that I have found is, unsurprisingly, [baseimage-alpine](https://github.com/linuxserver/docker-baseimage-alpine/pkgs/container/baseimage-alpine) at 21.7 MB. Alpine images are deliberately minimal.

.github/workflows/test.yaml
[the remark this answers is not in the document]
Kurt-von-Laven: It's quite subjective. I would put them first on the basis that the URL bears more relation to the headers and HTTP method than it does to the other arguments.

.github/workflows/test.yaml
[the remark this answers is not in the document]
Kurt-von-Laven: No, I mean literally just add a line that says `name: Notify` below this line.

action.yaml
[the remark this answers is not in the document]
Kurt-von-Laven: The conventions used in documentation vary wildly. The only general advice I can give is to try to follow the convention of the project or organization for the sake of consistency. We haven't made a practice of using the imperative in documentation, although I don't have any particular argument against it in general. Mirroring the rootless-docker docs is reasonable in our case.

.github/workflows/test.yaml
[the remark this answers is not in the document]
Kurt-von-Laven: I would think [Delete GitHub Actions caches for a repository (using a cache key)](https://docs.github.com/en/rest/actions/cache#delete-github-actions-caches-for-a-repository-using-a-cache-key) is the pertinent endpoint though?

.github/workflows/test.yaml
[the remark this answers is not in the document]
mwarres: Gotcha.  That makes sense.  Thanks for clarifying

Kurt-von-Laven (changes requested):

.github/workflows/test.yaml
```diff
@@ -38,10 +39,71 @@ jobs:
...
+      - name: Check out repository.
+        uses: actions/checkout@v3.0.2
+      - name: Load Docker images.
+        uses: ./
+        with:
+          key: docker-cache-test
+      - name: Verify Docker loaded empty image from cache.
+        run: |
+          description="$(docker inspect --format '{{ index .Config.Labels "description" }}' empty)"
+          [[ $description == "empty image" ]]
```
Kurt-von-Laven, line 71: This test and the step name need to be updated to reflect the fact that we are no longer building a Docker image.

```yaml
title="$(docker inspect --format '{{ index .Config.Labels "org.opencontainers.image.title" }}' ghcr.io/linuxserver/baseimage-alpine:edge-version-f502fcfd)"
[[ $title == "Baseimage-alpine" ]]
```

.github/workflows/test.yaml
```diff
@@ -38,10 +39,71 @@ jobs:
...
+    name: Delete Cache
+    if: needs.save-cache.result == 'success'
+    needs:
+      - save-cache
+      - restore-cache
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: write
+    steps:
+      - name: Delete existing test cache.
```
Kurt-von-Laven, line 82: nit: Remove "existing" here since it can be read to imply that we are about to create some other test cache.

.dictionary.txt
```diff
@@ -1 +1,3 @@
+baseimage
```
Kurt-von-Laven, line 1: Nit (commit message): Prefer one space after periods.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (changes requested):
For reasons I don't yet understand, the cache still has a number of unnecessary images and is now slightly larger at ~1380 MB.

.github/workflows/test.yaml
```diff
@@ -38,10 +39,64 @@ jobs:
...
+          key: docker-cache-test
+      - name: Pull a small test Docker image.
+        run: docker pull ghcr.io/linuxserver/baseimage-alpine:edge-version-f502fcfd
+  restore-cache:
+    name: Restore Cache
+    needs:
+      - save-cache
+    runs-on: ubuntu-20.04
+    permissions:
+      actions: write
```
Kurt-von-Laven, line 62: Recommend adding `# for cache deletion` so that we know when this permission might need to be removed in the future.

cspell.config.yaml
```diff
@@ -11,3 +11,4 @@ dictionaries:
   - fullstack
   - npm
   - python
+  - docker
```
Kurt-von-Laven, line 14: Nit: please maintain alphabetical order for consistency.

Kurt-von-Laven (changes requested):
Maybe we should just go ahead and add the filtering feature before merging this one since I still have no idea where all these extra images are coming from, and we really don't want to waste 1.3 GB of precious cache space.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM**

PR #88, opened 2022-07-22
test(integration): Change Jest to `resetMocks`

mwarres:
Change Jest configuration to `resetMocks` instead of `clearMocks` to eliminate the most substantial mechanism of cross-test interaction.

### DIFF
--- a/src/integration.test.ts
+++ b/src/integration.test.ts
@@ -18,18 +18,22 @@ describe("Integration Test", (): void => {
 
   let child_process: Mocked<typeof import("node:child_process")>;
   let cache: Mocked<typeof import("@actions/cache")>;
-  let inMemoryCache: Record<string, string>;
   let core: Mocked<typeof import("@actions/core")>;
-  let state: Record<string, string>;
   let util: typeof import("./util.js");
   let docker: typeof import("./docker.js");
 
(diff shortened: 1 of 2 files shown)


Kurt-von-Laven (changes requested):
Jest should be capitalized, and you need to move the mocking from `beforeAll` to `beforeEach` in the integration test so the mocks are configured after the mocks are reset.

src/integration.test.ts
```diff
@@ -24,7 +24,9 @@ describe("Integration Test", (): void => {
   let util: typeof import("./util.js");
   let docker: typeof import("./docker.js");
 
-  beforeAll(async (): Promise<void> => {
+  beforeEach(async (): Promise<void> => {
+    inMemoryCache = {};
+    state = {};
```
Kurt-von-Laven, line 29: Nit: recommend either moving these assignments below the imports with a blank line between (on the basis that imports tend to be at the top of files, so they could at least be at the top of the function in this case) or performing the assignments in the same order as the declarations (on the basis of convention).

Kurt-von-Laven (changes requested):
One minor comment, but this looks great.

Kurt-von-Laven (approved):
**LGTM**

PR #89, opened 2022-07-23
ci(GitHub-Actions): Grant minimum necessary scopes

Kurt-von-Laven:
Grant callable workflows the permissions they require. The Notify Assignee and Notify Reviewers workflows require the `contents:read` scope. [actions/checkout](https://github.com/actions/checkout) uses the `contents:read` scope to check out the calling repository. Granting this specific permission also has the effect of reducing the scope of all unspecified permissions from read to none. Since this repository is public, most of its data can be read without additional permissions, but the `contents:read` scope is needed in the callable workflows when the calling repository is private. Callable workflows can only be called when granted the permissions they require.


mwarres (approved):
LGTM

PR #98, opened 2022-08-02
perf(action): Filter out pre-cached images

### LINKED ISSUE #45: Restoring the cache takes longer than running docker pull
Edit: I know that the additional images were saved in the post action in the job that populated the cache, I do not know how they ended up in `docker image list` in that job.

The log from a cache hit on https://github.com/contiki-ng/contiki-ng/pull/1946 below.

`Cache Size:` is printed after 25 seconds, `tar` takes 59 seconds, and `docker load` takes 2 min and 14 seconds. The typical docker pull command without any cache takes less than 1 minute and 30 seconds.

The CI has an explicit step to pull the right version of `contiker/contiki-ng`, and it's possible that brings along the Ubuntu 18.04 image which it is based on. I am not sure where the other images come from.
```
Run ScribeMD/docker-cache@0.1.4
Received 192937984 of 3148759409 (6.1%), 183.8 MBs/sec
Received 423624704 of 3148759409 (13.5%), 201.6 MBs/sec
Received 650117120 of 3148759409 (20.6%), 206.4 MBs/sec
Received 876609536 of 3148759409 (27.8%), 208.7 MBs/sec
Received 1111490560 of 3148759409 (35.3%), 211.7 MBs/sec
Received 1354760192 of 3148759409 (43.0%), 214.9 MBs/sec
Received 1581252608 of 3148759409 (50.2%), 215.0 MBs/sec
Received 1820327936 of 3148759409 (57.8%), 216.4 MBs/sec
Received 2071986176 of 3148759409 (65.8%), 219.0 MBs/sec
Received 2147483647 of 3148759409 (68.2%), 115.8 MBs/sec
Received 2306867199 of 3148759409 (73.3%), 117.8 MBs/sec
Received 2462056447 of 3148759409 (78.2%), 119.3 MBs/sec
Received 2634022911 of 3148759409 (83.7%), 121.4 MBs/sec
Received 2822766591 of 3148759409 (89.6%), 124.1 MBs/sec
Received 3003121663 of 3148759409 (95.4%), 126.2 MBs/sec
Received 3148759409 of 3148759409 (100.0%), 116.5 MBs/sec
Cache Size: ~3003 MB (3148759409 B)
/usr/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/aa3ae13e-32e5-4393-878e-987f4e070a15/cache.tzst -P -C /home/runner/work/contiki-ng/contiki-ng
Cache restored successfully

docker load --input ~/.docker-images.tar
Loaded image: node:14
Loaded image: node:16
Loaded image: buildpack-deps:buster
Loaded image: node:14-alpine
Loaded image: alpine:3.13
Loaded image: debian:9
Loaded image: moby/buildkit:latest
Loaded image: ubuntu:20.04
Loaded image: ubuntu:18.04
Loaded image: alpine:3.12
Loaded image: contiker/contiki-ng:63f7e91e3
Loaded image: buildpack-deps:stretch
Loaded image: node:12
Loaded image: alpine:3.14
Loaded image: ubuntu:16.04
Loaded image: buildpack-deps:bullseye
Loaded image: debian:10
Loaded image: debian:11
Loaded image: node:16-alpine
Loaded image: node:12-alpine
```

mwarres:
Fixes #45.

Only save relevant Docker images. Improve performance when run as root, e.g., in Test workflow, the Save Cache job ran for approximately 5 minutes and now runs in less than 10 seconds. The Restore Cache job ran in approximately 40 seconds and now runs in less than 10 seconds. Previously, unnecessary Docker images were saved because GitHub Actions pre-caches a standard set of Docker images when Docker runs as root; those images aren't accessible when using rootless-docker, so this change doesn't impact users of rootless-docker.


src/docker.ts
```diff
@@ -26,13 +36,32 @@ const saveDockerImages = async (): Promise<void> => {
...
+      await execBashCommand(
+        'docker image list --format "{{ .Repository }}:{{ .Tag }}"'
+      )
+    ).split("\n");
+    console.log(`fullDockerImagesList = ${fullDockerImagesList}`);
+    const dockerImagesToBeSaved = fullDockerImagesList.filter(
+      (image) => !unneededDockerImagesList.includes(image)
     );
+    if (dockerImagesToBeSaved.length === 0) {
+      info("No docker images to save");
```
Kurt-von-Laven, line 50: Nit: "docker" --> "Docker"

src/integration.test.ts
```diff
@@ -15,6 +15,10 @@ const getKey = (paths: string[], key: string): string => {
 describe("Integration Test", (): void => {
   const KEY = "a-cache-key";
   const EXEC_OPTIONS = { shell: "/usr/bin/bash" };
+  const STDOUT = "standard output from Bash command";
+  const STDERR = "error output from Bash command";
+
+  const dockerImagesList: string[] = [];
```
Kurt-von-Laven, line 21: Nit: `dockerImagesList` --> `dockerImages`

src/docker.ts
```diff
@@ -26,13 +36,32 @@ const saveDockerImages = async (): Promise<void> => {
...
+    } else {
+      const formattedDockerImagesToBeSaved = dockerImagesToBeSaved.join(" ");
+      await execBashCommand(
+        `docker save --output ${DOCKER_IMAGES_PATH} ${formattedDockerImagesToBeSaved}`
+      );
+    }
     await saveCache([DOCKER_IMAGES_PATH], key);
```
Kurt-von-Laven, line 57: Why are we calling `saveCache` if there are no Docker images to save? This is a good opportunity for test-driven development: improve the tests so that the current production code fails, then correct the production code so that the improved tests pass.

src/integration.test.ts
```diff
@@ -58,51 +81,42 @@ describe("Integration Test", (): void => {
...
-      (_command: any, _options: any, callback: any): any => {
-        callback(null, { stdout, stderr });
-      }
-    );
-
-    await (load ? docker.loadDockerImages() : docker.saveDockerImages());
-
+  const execAsserts = (command: string): void => {
     expect(core.getInput).nthCalledWith(1, "key", { required: true });
-    expect(core.info).nthCalledWith<[string]>(1, command);
```
Kurt-von-Laven, line 73: I suspect it isn't necessary to remove this provided the correct `command` is passed.

src/integration.test.ts
```diff
@@ -58,51 +81,42 @@ describe("Integration Test", (): void => {
...
     );
-    expect(core.info).lastCalledWith(stdout);
-    expect(core.error).lastCalledWith(stderr);
+    expect(core.error).lastCalledWith(STDERR);
     expect(core.setFailed).not.toHaveBeenCalled();
   };
 
   test("cache misses, then hits", async (): Promise<void> => {
     // Attempt first cache restore.
+    const listCommand = `docker image list --format "{{ .Repository }}:{{ .Tag }}"`;
```
Kurt-von-Laven, line 97: Nit: prefer single quotes to back ticks for consistency since no string interpolation is being performed here.

Kurt-von-Laven (changes requested):
Good work! It's probably worth another pass with fresh eyes after addressing review feedback to double check that all of the proper assertions were actually added since tests don't fail by themselves.

src/docker.test.ts
```diff
@@ -46,6 +46,7 @@ describe("Docker images", (): void => {
```
Kurt-von-Laven, line 46: Update commit message to reflect the fact that GitHub Actions pre-caches a standard set of Docker images but that those aren't accessible when using rootless-docker. Also, "Docker" should be capitalized when it's a proper noun.

src/docker.ts
```diff
@@ -14,6 +15,15 @@ const loadDockerImages = async (): Promise<void> => {
...
   if (cacheHit) {
     await execBashCommand(`docker load --input ${DOCKER_IMAGES_PATH}`);
+  } else {
+    const dockerImagesList = await execBashCommand(
+      `docker image list --format "{{ .Repository }}:{{ .Tag }}"`
+    );
+    console.log(
+      `DOCKER_IMAGES_LIST = ${DOCKER_IMAGES_LIST}`,
+      `dockerImagesList = ${dockerImagesList}`
+    );
```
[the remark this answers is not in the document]
mwarres: Oops! Yep! Thanks for catching this!

src/integration.test.ts
```diff
@@ -33,6 +37,25 @@ describe("Integration Test", (): void => {
...
+      (_command: any, _options: any, callback: any): any => {
+        let stdout;
+        if (
+          _command !==
+          'docker image list --format "{{ .Repository }}:{{ .Tag }}"'
+        ) {
+          stdout = STDOUT;
+        } else {
+          callCount += 1;
+          dockerImagesList.push(`test-docker-image:v${callCount}`); // see saveDockerImages() in docker.ts.  The variable callCount ensures that fullDockerImagesList has more unique elements than unneededDockerImagesList; this is necessary to save docker images.
```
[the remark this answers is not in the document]
mwarres: I took a stab at this (still working on the rest of the comments so I haven't pushed it yet) but have have no idea if I'm achieving what you intended.  Will be interested in your feedback!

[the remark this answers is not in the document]
Kurt-von-Laven: Feel free to paste the next iteration if you want to workshop it since writing can sometimes benefit from more rounds of feedback than code.

src/docker.ts
mwarres: Oh wow, good point.  I didn't notice that!!

mwarres: I'm failing to find a reasonable way to implement this.  None of the current test cases in `docker.ts` seem to me to be obvious matches for the scenario where there are no Docker images to save.  Thus, I looked in the `integration.test.ts` and figured that it could go in the cache miss part of the integration test.  However, then I realized that with the current mocking situation, `child_process.exec.mockImplementation` is designed to increment the counter in the case where the command is `docker image list...` so that we never will get a situation where we are recreating a simulation of no Docker images to save.  I feel like I could maybe make a separate branch for the mock but this seems messy.  Am I missing something obvious/what is a better way of thinking of this??  Thanks in advance for your thoughts.

src/integration.test.ts
mwarres: This assert (specifically with n = 1) was appropriate when called after the `listCommand` and the `loadCommand` but not with the `saveCommand` (where n needs to be 3 in order for the test to pass).  I took this assert out of the `execAsserts` function and just wrote it into the code block of asserts following each command with the appropriate n.  Do you think that solution is ok?

src/docker.ts
Kurt-von-Laven: Oops, you are right. We need a new test case in `docker.test.ts` for this scenario.

src/integration.test.ts
Kurt-von-Laven: What I meant was that you could instead call `execAsserts('docker image list --format "{{ .Repository }}:{{ .Tag }}"')` and then assert that the third call to `core.info` is with `saveCommand` separately.

src/integration.test.ts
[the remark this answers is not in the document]
mwarres: // When docker images list is called and Docker is running in root, docker images list generates a list that includes a standard set of Docker images. In our test workflow, the goal is to only save an empty test Docker image instead of a comprehensive list (to minimize cache use). Generating a different Docker images list in the saving step vs. the loading step and only caching the difference between the two lists achieves this goal.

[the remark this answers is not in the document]
Kurt-von-Laven: "in" --> "as"

Recommend elaborating that the standard set of Docker images are pre-cached by GitHub Actions since otherwise it almost sounds like you could be describing a characteristic of Docker itself.

Keep in mind that this is the integration test, not the GitHub Actions test workflow, so the cache library is mocked out.

Here are some minor grammatical adjustments for the last sentence: "Generating a different Docker image list in the save step vs. the load step achieves this goal since the production code saves the difference between the two lists."

src/integration.test.ts
mwarres: I might be missing something, but I think that if I call `execAsserts('docker image list --format "{{ .Repository }}:{{ .Tag }}"')` that I will fail the following assert:

Kurt-von-Laven: Ah, I didn't consider that. Does it work if you replace `lastCalledWith` with `nthCalledWith` for some value of `n`?

mwarres: I ended up adding `n: number` as an argument to `execAsserts` and adding `expect(core.info).nthCalledWith<[string]>(n, command);` back into that function.  Then I am calling `execAsserts` with n = 1 for the `loadCommand` and `listCommand` scenarios and n = 3 for the `saveCommand` scenario.  It seems pretty simple and passes the test.  Let me know if that solution suits what you are hoping for!

Kurt-von-Laven: It is definitely a step forward, but I think it can be even simpler since I don't think it's necessary to pass `n`.

mwarres: I need to work on this one.  I'll do another force-push once I think I've gotten it.  In the meantime, I think/hope I addressed the rest of your feedback.

src/integration.test.ts
```diff
@@ -33,6 +37,25 @@ describe("Integration Test", (): void => {
     inMemoryCache = {};
     state = {};
 
+    let callCount = 0;
+
+    child_process.exec.mockImplementation(
+      (_command: any, _options: any, callback: any): any => {
+        let stdout;
+        if (
+          _command ==
```
Kurt-von-Laven, line 46: `==` --> `===` since [the former operator (loose equality) doesn't mean anything sensible in JavaScript](https://dorey.github.io/JavaScript-Equality-Table/).

src/integration.test.ts
```diff
@@ -33,6 +37,25 @@ describe("Integration Test", (): void => {
...
+          'docker image list --format "{{ .Repository }}:{{ .Tag }}"'
+        ) {
+          callCount += 1;
+          dockerImages.push(`test-docker-image:v${callCount}`); // When docker images list is called and Docker is running as root, docker images list generates a list that includes a standard set of Docker images.  The standard set of Docker images are pre-cached by GitHub Actions. In our test workflow, the goal is to only save an empty test Docker image instead of a comprehensive list (to minimize cache use). Generating a different Docker image list in the save step vs. the load step achieves this goal since the production code saves the difference between the two lists.
```
Kurt-von-Laven, line 50: - Prefer multi-line comments to horizontal scrolling since research has shown that reading speed is optimized by short line lengths.
- Remove "docker images list is called and" since it's redundant.
- Replace ".  The standard set of Docker images" with "that" for the sake of conciseness.
- > Generating a different Docker image list in the save step vs. the load step achieves this goal
  
  This bit can be read to mean that the test code achieves the aforementioned goal of the action itself. Sorry I neglected to mention that the following piece of feedback was intended to apply to this sentence as well after the grammatical issues were fixed:
  
  > Keep in mind that this is the integration test, not the GitHub Actions test workflow [...]

Kurt-von-Laven (changes requested):
Lots of great progress here. Good to see more asserts working their way in. This is getting close to landing.

src/integration.test.ts
mwarres: Thanks for the feedback and thanks for reinforcing the last bit where  you restate "Keep in mind that this is the integration test, not the GitHub Actions test workflow [...]"  I definitely didn't fully understand what you meant previously.  I think I do now but if my updated comment is off, then I think I'm still missing the required conceptual understanding to get this correct.

mwarres: Note: I'm pushing the changes with the latest updated comments soon but haven't yet.

src/docker.test.ts
```diff
@@ -158,4 +191,9 @@ describe("Docker images", (): void => {
 
     assertCacheHitSave();
   });
+
+  test("aren't saved when new docker image list is empty", async (): Promise<void> => {
+    await mockedSaveDockerImages(false, false, true);
+    expect(cache.saveCache).not.toHaveBeenCalled();
+  });
```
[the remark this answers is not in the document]
mwarres: I added quite a few asserts--let me know if what I've added is satisfactory or if I should add more!

src/integration.test.ts
```diff
@@ -58,58 +88,75 @@ describe("Integration Test", (): void => {
...
   };
 
+  const assertLoadDockerImages = (cacheHit: boolean): void => {
+    expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
+    expect(core.saveState).nthCalledWith(1, docker.CACHE_HIT, cacheHit);
+    expect(core.setOutput).nthCalledWith(1, docker.CACHE_HIT, cacheHit);
+    if (!cacheHit) {
+      expect(core.saveState).lastCalledWith(
+        docker.DOCKER_IMAGES_LIST,
+        expect.anything()
```
Kurt-von-Laven, line 111: We can be more specific here.

src/integration.test.ts
```diff
@@ -58,58 +88,75 @@ describe("Integration Test", (): void => {
...
+    const saveCommand = `docker save --output ${docker.DOCKER_IMAGES_PATH} test-docker-image:v2`;
+    await docker.saveDockerImages();
+    expect(core.getState).nthCalledWith(1, docker.CACHE_HIT);
+    expect(core.getInput).lastCalledWith("read-only");
+    expect(core.getState).lastCalledWith(docker.DOCKER_IMAGES_LIST);
+    expect(cache.saveCache).toHaveBeenCalledTimes(1);
+    expect(cache.saveCache).nthCalledWith(
+      1,
+      [docker.DOCKER_IMAGES_PATH],
+      expect.anything()
```
Kurt-von-Laven, line 144: We can be more specific here.

src/integration.test.ts
```diff
@@ -58,58 +88,75 @@ describe("Integration Test", (): void => {
...
-      '2>&1 xargs --delimiter="\n" --no-run-if-empty --verbose --exit ' +
-      `docker save --output ${docker.DOCKER_IMAGES_PATH}`;
-    await mockedExec(false, saveCommand);
+    const saveCommand = `docker save --output ${docker.DOCKER_IMAGES_PATH} test-docker-image:v2`;
+    await docker.saveDockerImages();
+    expect(core.getState).nthCalledWith(1, docker.CACHE_HIT);
+    expect(core.getInput).lastCalledWith("read-only");
+    expect(core.getState).lastCalledWith(docker.DOCKER_IMAGES_LIST);
+    expect(cache.saveCache).toHaveBeenCalledTimes(1);
+    expect(cache.saveCache).nthCalledWith(
```
Kurt-von-Laven, line 141: Prefer `lastCalledWith` where possible for simplicity? `nthCalledWith` is a little like introducing yourself with your middle name.

src/integration.test.ts
```diff
@@ -58,58 +88,75 @@ describe("Integration Test", (): void => {
...
-      'docker image list --format "{{ .Repository }}:{{ .Tag }}" | ' +
-      '2>&1 xargs --delimiter="\n" --no-run-if-empty --verbose --exit ' +
-      `docker save --output ${docker.DOCKER_IMAGES_PATH}`;
-    await mockedExec(false, saveCommand);
+    const saveCommand = `docker save --output ${docker.DOCKER_IMAGES_PATH} test-docker-image:v2`;
+    await docker.saveDockerImages();
+    expect(core.getState).nthCalledWith(1, docker.CACHE_HIT);
+    expect(core.getInput).lastCalledWith("read-only");
+    expect(core.getState).lastCalledWith(docker.DOCKER_IMAGES_LIST);
+    expect(cache.saveCache).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 140: In theory, this could be asserted about almost everything, but that seems like overkill. Is there a particular reason to use it here.

src/integration.test.ts
Kurt-von-Laven, line 111: We can be more specific here.

Kurt-von-Laven (changes requested):
Good progress!

src/docker.ts
```diff
@@ -26,13 +32,31 @@ const saveDockerImages = async (): Promise<void> => {
...
-        '2>&1 xargs --delimiter="\n" --no-run-if-empty --verbose --exit ' +
-        `docker save --output ${DOCKER_IMAGES_PATH}`
+    const preexistingDockerImages = getState(DOCKER_IMAGES_LIST).split("\n");
+    const dockerImages = await execBashCommand(
+      'docker image list --format "{{ .Repository }}:{{ .Tag }}"'
     );
-    await saveCache([DOCKER_IMAGES_PATH], key);
+    const dockerImagesList = dockerImages.split("\n");
+    const newDockerImages = dockerImagesList.filter(
+      (image) => !preexistingDockerImages.includes(image)
```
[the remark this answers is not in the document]
mwarres: I have fixed it both times, although the first time I didn't trust my memory and assumed I had forgotten to do it.  I think one of the linters is changing it after I fix it.  Could I have a setting somehow that is different than it should be?

src/integration.test.ts
```diff
@@ -58,51 +81,40 @@ describe("Integration Test", (): void => {
...
-    expect(core.error).lastCalledWith(stderr);
+    expect(core.info).nthCalledWith<[string]>(n, command);
+    expect(core.error).lastCalledWith(STDERR);
     expect(core.setFailed).not.toHaveBeenCalled();
   };
 
   test("cache misses, then hits", async (): Promise<void> => {
     // Attempt first cache restore.
+    const listCommand =
+      'docker image list --format "{{ .Repository }}:{{ .Tag }}"';
```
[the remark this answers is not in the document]
mwarres: I also fixed this one.  I think a linter is changing it back.  Let me know if there is a setting or something I can change so that this doesn't happen.

src/docker.test.ts
```diff
@@ -46,27 +46,51 @@ describe("Docker images", (): void => {
...
     if (!cacheHit) {
       core.getInput.mockReturnValueOnce(readOnly.toString());
     }
+    if (!readOnly && !emptyImageList) {
+      core.getState.mockReturnValueOnce("");
+      util.execBashCommand.mockResolvedValueOnce("test\ntest");
+    } else if (!readOnly) {
+      core.getState.mockReturnValueOnce("");
+      util.execBashCommand.mockResolvedValueOnce("");
+    }
```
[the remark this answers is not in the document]
mwarres: I fixed the `getState` call to return an actual list to exercise the subtraction.  In doing so, I created variables to represent the preexisting images and the new images.  When I tried to incorporate your feedback from the comment below and replace `emptyImageList: boolean` with `commandOutput: string`, I got the following TypeScript error: `A parameter initializer is only allowed in a function or constructor implementation.`  How would you like me to proceed in light of that?  I'm tempted to revert to `emptyImageList: boolean`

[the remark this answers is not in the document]
Kurt-von-Laven: It sounds like there was something else going on since there's no prohibition against string arguments.

src/integration.test.ts
```diff
@@ -15,6 +15,12 @@ const getKey = (paths: string[], key: string): string => {
 describe("Integration Test", (): void => {
   const KEY = "a-cache-key";
   const EXEC_OPTIONS = { shell: "/usr/bin/bash" };
+  const STDOUT = "standard output from Bash command";
+  const STDERR = "error output from Bash command";
+  const listCommand =
```
Kurt-von-Laven, line 20: Constants are `ALL_CAPS`.

src/integration.test.ts
```diff
@@ -32,6 +39,34 @@ describe("Integration Test", (): void => {
...
+    loadCommand = `docker load --input ${docker.DOCKER_IMAGES_PATH}`;
+
+    let callCount = 0;
+
+    child_process.exec.mockImplementation(
+      (_command: any, _options: any, callback: any): any => {
+        let stdout;
+        if (
+          _command ===
+          'docker image list --format "{{ .Repository }}:{{ .Tag }}"'
```
Kurt-von-Laven, line 51: You can use the constant here.

src/integration.test.ts
```diff
@@ -32,6 +39,34 @@ describe("Integration Test", (): void => {
...
+           * difference between the restore and save steps so there is something to save.
+           */
+          callCount += 1;
+          dockerImages.push(`test-docker-image:v${callCount}`);
+          stdout = dockerImages.join("\n");
+        } else {
+          stdout = STDOUT;
+        }
+        callback(null, { stdout, stderr: STDERR });
+        expect(core.info).lastCalledWith(stdout);
```
Kurt-von-Laven, line 67: This is racy; there's no guarantee that it executes after `util.execAsPromised` returns. The race condition can be eliminated by moving this line to `assertExecBashCommand`.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
     expect(core.getInput).nthCalledWith(1, "key", { required: true });
+    assertExecBashCommand(1, command);
+  };
+
+  const assertLoadCacheMiss = (): void => {
+    expect(core.saveState).lastCalledWith(
+      docker.DOCKER_IMAGES_LIST,
+      "test-docker-image:v1"
+    );
+    expect(core.info).lastCalledWith(dockerImages.join("\n"));
```
Kurt-von-Laven, line 119: This is duplicitous with the assert currently in the mock implementation of `util.exec`.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
     expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, false);
```
Kurt-von-Laven, line 120: This is duplicitous with the assert in `assertLoadDockerImages`.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
-    // Attempt first cache restore.
-    await docker.loadDockerImages();
-
-    // Expect cache miss since cache has never been saved.
+  const assertListOrLoad = (command: string): void => {
     expect(core.getInput).nthCalledWith(1, "key", { required: true });
+    assertExecBashCommand(1, command);
+  };
+
+  const assertLoadCacheMiss = (): void => {
```
Kurt-von-Laven, line 114: Given how few lines will remain of this function, I would simply inline it.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
+    expect(core.saveState).lastCalledWith(
+      docker.DOCKER_IMAGES_LIST,
+      "test-docker-image:v1"
+    );
```
Kurt-von-Laven, line 118: Move to bottom of function to mirror production code.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
+    expect(core.error).lastCalledWith(STDERR);
     expect(core.setFailed).not.toHaveBeenCalled();
   };
 
-  test("cache misses, then hits", async (): Promise<void> => {
-    // Attempt first cache restore.
-    await docker.loadDockerImages();
-
-    // Expect cache miss since cache has never been saved.
+  const assertListOrLoad = (command: string): void => {
```
Kurt-von-Laven, line 109: This was a worthwhile experiment, but now that we have seen the result, I would simply inline this function given that it only saves a single line of duplication between all three call sites combined.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
+  };
+
+  const assertSaveCacheMiss = (): void => {
+    assertListOrLoad(listCommand);
+    expect(core.getInput).lastCalledWith("read-only");
+    expect(core.getState).lastCalledWith(docker.DOCKER_IMAGES_LIST);
+    expect(cache.saveCache).lastCalledWith(
+      [docker.DOCKER_IMAGES_PATH],
+      "a-cache-key"
+    );
```
Kurt-von-Laven, line 151: Move to bottom of function to mirror production code.

src/integration.test.ts
```diff
@@ -58,65 +93,93 @@ describe("Integration Test", (): void => {
...
+  const assertSaveDockerImages = (cacheHit: boolean): void => {
+    expect(core.getState).nthCalledWith(1, docker.CACHE_HIT);
+    if (cacheHit) {
+      assertSaveCacheHit();
+    } else {
+      assertSaveCacheMiss();
+    }
+  };
+
+  test("cache misses, then hits", async (): Promise<void> => {
```
Kurt-von-Laven, line 165: Wow, the top level integration test never looked half as good.

Kurt-von-Laven (changes requested):

mwarres (commented):
Note: I haven't divided the PR into separate PRs yet but will do so after the other issues with this PR are addressed satisfactorily.

src/integration.test.ts
```diff
@@ -58,65 +89,82 @@ describe("Integration Test", (): void => {
...
+  const assertLoadDockerImages = (cacheHit: boolean): void => {
+    expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
+    expect(core.saveState).nthCalledWith(1, docker.CACHE_HIT, cacheHit);
+    expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    if (cacheHit) {
+      expect(core.getInput).nthCalledWith(1, "key", { required: true });
+      assertExecBashCommand(1, loadCommand);
+    } else {
+      expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, false);
+      expect(core.getInput).nthCalledWith(1, "key", { required: true });
```
Kurt-von-Laven, line 117: Extricate duplicated assert from branch.

src/integration.test.ts
```diff
@@ -58,65 +89,82 @@ describe("Integration Test", (): void => {
...
+    expect(core.getInput).lastCalledWith("key", { required: true });
+    expect(core.info).lastCalledWith(
+      `Cache hit occurred on the primary key ${KEY}, not saving cache.`
+    );
     expect(child_process.exec).not.toHaveBeenCalled();
     expect(core.setFailed).not.toHaveBeenCalled();
+  };
+
+  const assertSaveCacheMiss = (): void => {
+    expect(core.getInput).nthCalledWith(1, "key", { required: true });
```
Kurt-von-Laven, line 136: This can be done once in `assertSaveDockerImages` rather than duplicated across branches.

src/docker.test.ts
```diff
@@ -46,27 +53,57 @@ describe("Docker images", (): void => {
...
     core.getState.mockReturnValueOnce(cacheHit.toString());
     if (!cacheHit) {
       core.getInput.mockReturnValueOnce(readOnly.toString());
+      if (!readOnly) {
+        core.getState.mockReturnValueOnce(PRE_EXISTING_IMAGES);
+        util.execBashCommand.mockResolvedValueOnce(stdout);
+      }
     }
     await docker.saveDockerImages();
-
```
Kurt-von-Laven, line 64: Nit: undo remove of blank line.

src/docker.test.ts
```diff
@@ -27,6 +27,13 @@ const assertCalledInOrder = (...mocks: jest.Mock[]): void => {
 
 describe("Docker images", (): void => {
   const KEY = "a-cache-key";
+  const imagesList = Array.from(
+    { length: 4 },
+    (_v, i) => `test-docker-image:v${i}`
+  );
+  const IMAGES = imagesList.join("\n");
+  const PRE_EXISTING_IMAGES = imagesList.slice(1, 3).reverse().join("\n");
```
Kurt-von-Laven, line 35: Nit: preexisting is one word.

src/docker.test.ts
```diff
@@ -158,4 +192,11 @@ describe("Docker images", (): void => {
 
     assertCacheHitSave();
   });
+
+  test("aren't saved on cache miss when new Docker image list is empty", async (): Promise<void> => {
+    await mockedSaveDockerImages(false, false, PRE_EXISTING_IMAGES);
+    expect(util.execBashCommand).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 198: Nit: we have been placing a blank line between the act and assert stanzas.

Kurt-von-Laven (changes requested):

src/docker.test.ts
```diff
@@ -109,18 +151,19 @@ describe("Docker images", (): void => {
   test("aren't loaded on cache miss", async (): Promise<void> => {
     await mockedLoadDockerImages(false);
 
-    expect(core.saveState).lastCalledWith(docker.CACHE_HIT, false);
-    expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, false);
-    expect(util.execBashCommand).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
mwarres: I haven't figured this one out.  Could you explain why renaming is necessary given this change?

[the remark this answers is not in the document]
Kurt-von-Laven: The name isn’t technically wrong that we don’t load Docker images on cache miss, but there’s no assert that corresponds to this anymore. Instead, we now assert that preexisting images are recorded.

Kurt-von-Laven:
Considering quantifying the approximate performance improvement of this change on the test workflow in the commit message.

Kurt-von-Laven (changes requested):
Commit message: when documenting performance it's useful it say the approximate runtime before and after the change since there's a huge difference between making a 1-hour job 1 minute faster and making a 2-minute job 1 minute faster. Consider "in test workflow: the" --> "in the test workflow, the" (or "in Test workflow, the").

src/docker.test.ts
```diff
@@ -27,6 +27,13 @@ const assertCalledInOrder = (...mocks: jest.Mock[]): void => {
```
Kurt-von-Laven, line 27: Commit message: "test workflow" --> "Test workflow" since it's a proper noun. I just spot checked a few of the Save Cache jobs, and I think it was more like 6-10 minutes previously, but I didn't investigate comprehensively. Remember to update the pull request description.

src/docker.test.ts
```diff
@@ -158,4 +197,12 @@ describe("Docker images", (): void => {
 
     assertCacheHitSave();
   });
+
+  test("aren't saved on cache miss when new Docker image list is empty", async (): Promise<void> => {
+    await mockedSaveDockerImages(false, false, PREEXISTING_IMAGES);
+
+    expect(util.execBashCommand).toHaveBeenCalledTimes(1);
+    expect(core.info).toHaveBeenCalledWith("No Docker images to save");
```
Kurt-von-Laven, line 205: Nit: prefer `lastCalledWith` for consistency.

Kurt-von-Laven (changes requested):

src/docker.test.ts
Kurt-von-Laven: "test workflow" --> "Test workflow"

Kurt-von-Laven:
Looks great! Let's decompose the portions that aren't related to the commit message into separate pre-cursor pull requests, and then rebase on those.

src/docker.ts
```diff
@@ -26,13 +32,30 @@ const saveDockerImages = async (): Promise<void> => {
...
     );
   } else {
-    await execBashCommand(
-      'docker image list --format "{{ .Repository }}:{{ .Tag }}" | ' +
-        '2>&1 xargs --delimiter="\n" --no-run-if-empty --verbose --exit ' +
-        `docker save --output ${DOCKER_IMAGES_PATH}`
+    const preexistingImages = getState(DOCKER_IMAGES_LIST).split("\n");
+    const images = await execBashCommand(
+      'docker image list --format "{{ .Repository }}:{{ .Tag }}"'
     );
```
Kurt-von-Laven, line 38: The logs read clearly to those who are acquainted with the internals of docker-cache, but look at the corresponding log output of the Save Cache job, and I think you will see they are somewhat unwelcoming to newcomers. It would help to offer some context on why we are executing this command both here and in `loadDockerImages`.

Kurt-von-Laven (changes requested):

src/docker.ts
```diff
@@ -26,13 +32,30 @@ const saveDockerImages = async (): Promise<void> => {
...
-    await saveCache([DOCKER_IMAGES_PATH], key);
+    const imagesList = images.split("\n");
+    const newImages = imagesList.filter(
+      (image) => !preexistingImages.includes(image)
```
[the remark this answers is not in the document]
mwarres: I'm not confident about the syntax.  I think this might be what you mean, but can you confirm?!
```
const newImages: string[] = imagesList.filter(
      (image: string) => !preexistingImages.includes(image)
    );
```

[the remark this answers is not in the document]
Kurt-von-Laven: TypeScript should be able to infer the type of `newImages`. I meant to specify the return type of the callback function like you have elsewhere, so you should only need to modify line 41.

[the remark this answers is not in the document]
Kurt-von-Laven: There is no need to specify the type of `newImages`.

Kurt-von-Laven (changes requested):

src/integration.test.ts
```diff
@@ -58,65 +89,88 @@ describe("Integration Test", (): void => {
...
       EXEC_OPTIONS,
       expect.anything()
     );
-    expect(core.info).lastCalledWith(stdout);
-    expect(core.error).lastCalledWith(stderr);
+    expect(core.error).lastCalledWith(STDERR);
     expect(core.setFailed).not.toHaveBeenCalled();
+    if (command === LIST_COMMAND) {
+      expect(core.info).nthCalledWith(3, dockerImages.join("\n"));
+    }
```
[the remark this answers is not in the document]
mwarres: Good point.  I think the most appropriate higher level placement is within `assertLoadDockerImages`.  I moved it there; let me know if you intended otherwise.

src/docker.ts
```diff
@@ -14,6 +15,14 @@ const loadDockerImages = async (): Promise<void> => {
   setOutput(CACHE_HIT, cacheHit);
   if (cacheHit) {
     await execBashCommand(`docker load --input ${DOCKER_IMAGES_PATH}`);
+  } else {
+    info(
+      "Recording preexisting Docker images. These include standard images pre-cached by GitHub Actions when Docker is run as root."
```
Kurt-von-Laven, line 20: Multiple instances: break long lines.

src/integration.test.ts
```diff
@@ -58,65 +89,95 @@ describe("Integration Test", (): void => {
...
+  const assertSaveCacheHit = (): void => {
+    expect(core.info).lastCalledWith(
+      `Cache hit occurred on the primary key ${KEY}, not saving cache.`
+    );
     expect(child_process.exec).not.toHaveBeenCalled();
     expect(core.setFailed).not.toHaveBeenCalled();
+  };
+
+  const assertSaveCacheMiss = (): void => {
+    const saveCommand = `docker save --output ${docker.DOCKER_IMAGES_PATH} test-docker-image:v2`;
```
Kurt-von-Laven, line 143: Recommend defining this immediately before it is used.

src/integration.test.ts
```diff
@@ -58,65 +89,95 @@ describe("Integration Test", (): void => {
...
+
```
Kurt-von-Laven, line 144: Nit: remove blank line.

src/integration.test.ts
```diff
@@ -58,65 +89,95 @@ describe("Integration Test", (): void => {
...
+    /* jscpd:ignore-end */
+    if (cacheHit) {
+      assertExecBashCommand(1, 1, loadCommand);
+    } else {
+      expect(core.info).nthCalledWith(
+        1,
+        "Recording preexisting Docker images. These include standard images pre-cached by GitHub Actions when Docker is run as root."
+      );
+      assertExecBashCommand(2, 1, LIST_COMMAND);
+      expect(core.info).nthCalledWith(3, dockerImages.join("\n"));
```
Kurt-von-Laven, line 126: It seems like the corresponding assert is missing after all other calls to `assertExecBashCommand`.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (changes requested):
Looks great! The unrelated changes just need to be separated out at this point.

src/docker.test.ts
```diff
@@ -47,32 +54,66 @@ describe("Docker images", (): void => {
   const assertLoadDockerImages = (cacheHit: boolean): void => {
     expect(core.getInput).lastCalledWith("key", { required: true });
     expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
-    expect(core.saveState).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    expect(core.saveState).nthCalledWith(1, docker.CACHE_HIT, cacheHit);
```
Kurt-von-Laven, line 57: Multiple instances: manually enforce type correctness on `nthCalledWith` calls within reason. You can see which cases we currently type and which we don't on main.

Kurt-von-Laven (changes requested):

src/docker.test.ts
```diff
@@ -47,32 +54,66 @@ describe("Docker images", (): void => {
...
     cacheHit: boolean,
     readOnly: boolean = false
+  ): void => {
+    expect(core.getInput).nthCalledWith(1, "key", { required: true });
+    expect(core.getState).nthCalledWith(1, docker.CACHE_HIT);
+    if (!cacheHit) {
+      expect(core.getInput).lastCalledWith("read-only");
+      if (!readOnly) {
+        expect(core.getState).lastCalledWith(docker.DOCKER_IMAGES_LIST);
+        expect(core.info).nthCalledWith(1, "Listing Docker images.");
```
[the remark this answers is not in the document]
Kurt-von-Laven: Whoops; good catch! Yeah, let's leave it in the helper function in that case.

Kurt-von-Laven (changes requested):

src/docker.test.ts
```diff
@@ -28,6 +28,13 @@ const assertCalledInOrder = (...mocks: jest.Mock[]): void => {
```
Kurt-von-Laven, line 28: The commit message was duplicated.

Kurt-von-Laven (changes requested):

src/integration.test.ts
```diff
@@ -59,69 +91,110 @@ describe("Integration Test", (): void => {
...
+    /* jscpd:ignore-start */
     expect(core.getInput).nthCalledWith<[string, InputOptions]>(1, "key", {
       required: true,
     });
-    expect(core.info).nthCalledWith<[string]>(1, command);
-    expect(child_process.exec).lastCalledWith(
-      command,
-      EXEC_OPTIONS,
-      expect.anything()
+    expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
```
Kurt-von-Laven, line 120: Multiple instances: this was probably my own gaffe as a reviewer, but it just hit me that there are a lot of unnecessary asserts in the integration test associated with the mock implementations. I can see the point of asserting that a mock implementation wasn't called (or, in certain cases, that it was called `n` times), but by and large asserting what it was called with will be redundant assuming the mock implementation ensures that the test will fail when the mocks are called with the wrong arguments.

Kurt-von-Laven (changes requested):

src/docker.test.ts
```diff
@@ -90,17 +131,17 @@ describe("Docker images", (): void => {
...
+  });
+
   test("exports DOCKER_IMAGES_PATH", (): void => {
     expect(docker.DOCKER_IMAGES_PATH).toBe("~/.docker-images.tar");
   });
 
   test("are loaded on cache hit", async (): Promise<void> => {
     await mockedLoadDockerImages(true);
 
-    expect(util.execBashCommand).lastCalledWith(
```
Kurt-von-Laven, line 100: Analogous to `"aren't saved on cache miss when new Docker image list is empty"`, it's probably worth asserting that we don't call `core.saveState` and that we only call `util.execBashCommand` once since we don't want to be needlessly listing the Docker images on cache hit.

src/util.test.ts
```diff
@@ -24,22 +24,23 @@ describe("Util", (): void => {
...
-      await util.execBashCommand(COMMAND);
+      const output = await util.execBashCommand(COMMAND);
 
       expect(core.info).nthCalledWith<[string]>(1, COMMAND);
       expect(child_process.exec).lastCalledWith(
         COMMAND,
         { shell: "/usr/bin/bash" },
         expect.anything()
       );
+      expect(output).toBe(stdout);
```
Kurt-von-Laven, line 43: When `error` is not `null`, we expect `output` to be `""`. The existing test enforces this expectation, but does it in something of a cryptic way spanning two levels of abstraction that will be more error-prone in the future.

Kurt-von-Laven (changes requested):

src/docker.test.ts
mwarres: Your feedback re: calling `util.execBashCommand` once makes sense to me.  However, I think that we do call `core.saveState` even if there is a cache hit in the production function.  Check this out (from `docker.ts`:
```const loadDockerImages = async (): Promise<void> => {
  const requestedKey = getInput("key", { required: true });
  const restoredKey = await restoreCache([DOCKER_IMAGES_PATH], requestedKey);
  const cacheHit = requestedKey === restoredKey;
  saveState(CACHE_HIT, cacheHit);```
Given that, do you still want me to make an assert re: `core.saveState`?  Perhaps asserting that is only called once? Let me know what you think makes sense or if I'm missing something.

Comment (Kurt-von-Laven) [reply src/docker.test.ts:100]: Good call; yeah, we should simply assert that it's called once.

Comment (mwarres) [reply src/docker.test.ts:100]: I think there were some function calls missing from the `assertCalledInOrder` call and that's why it was not clear that `core.saveState` is called within `loadDockerImages` even when there is a cache hit.  I fixed that and am tentatively pushing the result.

File: src/docker.test.ts [hunk 50]
```diff
@@ -90,42 +131,52 @@ describe("Docker images", (): void => {
...
-      `docker load --input ${docker.DOCKER_IMAGES_PATH}`
-    );
-
     /* The cache must be restored before the Docker images can be loaded. This
      * at least checks that the calls are made in the right order, but doesn't
      * ensure that the cache finished restoring before the Docker images started
      * loading.
      */
+    expect(core.saveState).toHaveBeenCalledTimes(1);
+    expect(util.execBashCommand).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 151: These two lines belong above the comment, which pertains to `assertCalledInOrder`. I would keep the blank line as well to avoid confusing Python developers, who may be accustomed to comments following rather than preceding what they describe.

src/util.test.ts
```diff
@@ -24,22 +24,24 @@ describe("Util", (): void => {
 
     const mockedExec = async (
       error: Error | null,
-      stdout: string | null = null,
-      stderr: string | null = null
+      stdout: string = "Stdout",
```
Kurt-von-Laven, line 27: This is a surprising default.

src/util.test.ts
```diff
@@ -24,22 +24,24 @@ describe("Util", (): void => {
...
-      await util.execBashCommand(COMMAND);
+      const output = await util.execBashCommand(COMMAND);
 
       expect(core.info).nthCalledWith<[string]>(1, COMMAND);
       expect(child_process.exec).lastCalledWith(
         COMMAND,
         { shell: "/usr/bin/bash" },
         expect.anything()
       );
+      if (error) expect(output).toBe(stderr);
```
Kurt-von-Laven, line 43: This is only passing by coincidence; we expect `output` to be `""` when `error` is not `null`. It will probably be easier to move the asserts on `output` to the top level to avoid branching.

Kurt-von-Laven (changes requested):

src/docker.test.ts
mwarres: I can't think of any plausible counterexamples; I'll delete those function calls.

src/docker.test.ts
```diff
@@ -90,16 +131,19 @@ describe("Docker images", (): void => {
...
   });
 
   test("are loaded on cache hit", async (): Promise<void> => {
     await mockedLoadDockerImages(true);
 
-    expect(util.execBashCommand).lastCalledWith(
-      `docker load --input ${docker.DOCKER_IMAGES_PATH}`
-    );
+    expect(core.saveState).toHaveBeenCalledTimes(1);
+    expect(util.execBashCommand).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 146: We can actually move this one assert to `assertLoadDockerImages` and perform it unconditionally.

Kurt-von-Laven (changes requested):

src/integration.test.ts
```diff
@@ -59,69 +90,87 @@ describe("Integration Test", (): void => {
...
       required: true,
     });
-    expect(core.info).nthCalledWith<[string]>(1, command);
-    expect(child_process.exec).lastCalledWith(
-      command,
-      EXEC_OPTIONS,
-      expect.anything()
+    expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    if (cacheHit) {
+      assertExecBashCommand(1, 1, loadCommand, STDOUT);
```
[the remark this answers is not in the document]
mwarres: I added an assert in the `if (cacheHit)` case testing that `core.saveState` is called exactly once.  Is it also appropriate to add an assert in the `else` case (i.e. the if there is not a cache hit) that asserts that `core.saveState` was called exactly twice?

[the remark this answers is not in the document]
Kurt-von-Laven: I wouldn't bother to assert that `core.saveState` was called exactly twice personally. I can see the argument for it, but it would imply adding a whole bunch of similar asserts elsewhere, and I don't believe it's very likely to catch a regression.

src/docker.test.ts
```diff
@@ -48,34 +55,69 @@ describe("Docker images", (): void => {
...
     expect(core.getInput).lastCalledWith("key", { required: true });
     expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
-    expect(core.saveState).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    expect(core.saveState).nthCalledWith<[string, boolean]>(
+      1,
+      docker.CACHE_HIT,
+      cacheHit
+    );
     expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    expect(util.execBashCommand).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 64: Let's assert the call count after the corresponding called withs for consistency with the rest of the tests and because the resulting error message is less specific when the call count is off than when the arguments don't match.

src/integration.test.ts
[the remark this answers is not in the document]
Kurt-von-Laven: We should assert that `child_process.exec` is called exactly once on cache hit in a pre-cursor pull request as well to keep this change as narrowly focused as possible.

src/integration.test.ts
```diff
@@ -59,69 +89,89 @@ describe("Integration Test", (): void => {
...
     });
-    expect(core.info).nthCalledWith<[string]>(1, command);
-    expect(child_process.exec).lastCalledWith(
-      command,
-      EXEC_OPTIONS,
-      expect.anything()
+    expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    expect(child_process.exec).toHaveBeenCalledTimes(1);
+    if (cacheHit) {
+      expect(core.saveState).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 117: Move this line below the following one for parity with the production code.

src/integration.test.ts
```diff
@@ -59,69 +89,89 @@ describe("Integration Test", (): void => {
...
     expect(core.getInput).nthCalledWith<[string, InputOptions]>(1, "key", {
       required: true,
     });
-    expect(core.info).nthCalledWith<[string]>(1, command);
-    expect(child_process.exec).lastCalledWith(
-      command,
-      EXEC_OPTIONS,
-      expect.anything()
+    expect(core.setOutput).lastCalledWith(docker.CACHE_HIT, cacheHit);
+    expect(child_process.exec).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 115: Move this line after the branch so that the corresponding called withs come first.

src/integration.test.ts
```diff
@@ -59,69 +90,87 @@ describe("Integration Test", (): void => {
...
+  };
+
+  const assertSaveCacheHit = (): void => {
+    expect(core.info).lastCalledWith(
+      `Cache hit occurred on the primary key ${KEY}, not saving cache.`
     );
-    expect(core.info).lastCalledWith(stdout);
-    expect(core.error).lastCalledWith(stderr);
+    expect(child_process.exec).not.toHaveBeenCalled();
     expect(core.setFailed).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
Kurt-von-Laven: This was added on `main` but needs to be added on this branch as well.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM** No further questions your honor. Congratulations on finishing this one!

PR #99, opened 2022-08-02
ci(GitHub-Actions): Improve pre-commit job name

mwarres:
Change job name to keep naming convention consistent with other repos.


Kurt-von-Laven (approved):
**LGTM** Thank you!

PR #101, opened 2022-08-03
test(docker): Decompose duplicate asserts

mwarres:
Group duplicated asserts in a helper function.

### DIFF
--- a/src/docker.test.ts
+++ b/src/docker.test.ts
@@ -69,6 +69,14 @@ describe("Docker images", (): void => {
     expect(core.getState).lastCalledWith(docker.CACHE_HIT);
   };
 
+  const assertCacheHitSave = (): void => {
+    expect(core.info).lastCalledWith(
+      `Cache hit occurred on the primary key ${KEY}, not saving cache.`
+    );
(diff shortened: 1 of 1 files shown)


src/docker.test.ts
```diff
@@ -69,6 +69,14 @@ describe("Docker images", (): void => {
     expect(core.getState).lastCalledWith(docker.CACHE_HIT);
   };
 
+  const assertsCacheHitSave = (): void => {
```
Kurt-von-Laven, line 72: Nit: usually verbs or verb phrases are preferred for function names, which in this case can be achieved by removing the first "s."

Kurt-von-Laven (changes requested):
Commitizen bumps the patch version numbers for redactors, so they are intended exclusively for refactoring of production code. I would use `test(docker)` for this instead even though you are only refactoring. "Decompose" would be a more appropriate term than "Remove" here, which conveys deletion.

Kurt-von-Laven (approved):
**LGTM**

Kurt-von-Laven:
I updated the pull request title to match the new commit message. Good choice of positioning within the file for the helper function!

PR #106, opened 2022-08-09
docs(VSCode): Recommend Python extension

mwarres:
The Python extension automatically activates the Poetry shell. We use Poetry to install pre-commit and Commitizen.


Kurt-von-Laven (approved):
**LGTM**

PR #110, opened 2022-08-18
test(docker): Improve test modularity

mwarres:
Introduce `assertLoadDockerImages` to call repeated asserts. Add `getInput` argument to `assertCalledInOrder` to test `readOnly` feature.

### DIFF
--- a/src/docker.test.ts
+++ b/src/docker.test.ts
@@ -44,12 +44,18 @@ describe("Docker images", (): void => {
     core.getInput.mockReturnValueOnce(KEY);
   });
 
+  const assertLoadDockerImages = (cacheHit: boolean): void => {
+    expect(core.getInput).lastCalledWith("key", { required: true });
+    expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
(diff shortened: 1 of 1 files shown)


Kurt-von-Laven (changes requested):
I would drop the last commit as it's currently broken in a subjectively more readable way.

src/docker.test.ts
```diff
@@ -44,12 +44,18 @@ describe("Docker images", (): void => {
     core.getInput.mockReturnValueOnce(KEY);
   });
 
+  const assertLoadDockerImages = (cacheHit: boolean): void => {
+    expect(core.getInput).lastCalledWith("key", { required: true });
+    expect(cache.restoreCache).lastCalledWith([docker.DOCKER_IMAGES_PATH], KEY);
+    expect(core.saveState).nthCalledWith(1, docker.CACHE_HIT, cacheHit);
```
Kurt-von-Laven, line 50: This should be `lastCalledWith` on this branch.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM**

PR #112, opened 2022-08-18
ci(dictionary): Add "deps" to custom dictionary

mwarres:
Dependabot uses "deps," short for "dependencies," in commit messages, which Commitizen incorporates into the changelog.


Kurt-von-Laven (approved):
**LGTM**

PR #114, opened 2022-08-23
chore(Poetry): Upgrade from 1.1.14 to 1.1.15

Kurt-von-Laven:
`poetry lock --no-update` did not produce consistent results locally and in CI, so the `poetry-lock` pre-commit hook failed in CI. Pull in a fix that consistently sorts the `package.extras` sections of the Poetry lock file.


mwarres (approved):
LGTM!

PR #121, opened 2022-08-25
test(action): Type nthCalledWith asserts

mwarres:


src/docker.test.ts
```diff
@@ -1,3 +1,4 @@
+import type { InputOptions } from "@actions/core";
```
Kurt-von-Laven, line 1: Multiple instances: let's put this right above the other `import type`.

src/docker.test.ts
```diff
@@ -68,7 +69,9 @@ describe("Docker images", (): void => {
     }
     await docker.saveDockerImages();
 
-    expect(core.getInput).nthCalledWith(1, "key", { required: true });
+    expect(core.getInput).nthCalledWith<[string, InputOptions]>(1, "key", {
```
Kurt-von-Laven, line 72: Multiple instances: can we fit all of the arguments on the line below, or does Prettier force this awkward formatting?

Kurt-von-Laven (changes requested):

src/docker.test.ts
mwarres: I attempted your formatting suggestion but, alas, Prettier objected.

Kurt-von-Laven (approved):
**LGTM**

PR #127, opened 2022-08-29
docs(README): Prefer three-digit shorthand hex

Kurt-von-Laven:
Simplify `#000000` to #000 and `#339933` to #393 in badge colors.


mwarres (approved):
LGTM

PR #128, opened 2022-08-29
test(docker): Reorder asserts

mwarres:
Reorder asserts to match production code.

### DIFF
--- a/src/docker.test.ts
+++ b/src/docker.test.ts
@@ -72,10 +72,10 @@ describe("Docker images", (): void => {
     expect(core.getInput).nthCalledWith<[string, InputOptions]>(1, "key", {
       required: true,
     });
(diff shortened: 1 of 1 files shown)


src/docker.test.ts
```diff
@@ -72,10 +72,10 @@ describe("Docker images", (): void => {
     expect(core.getInput).nthCalledWith<[string, InputOptions]>(1, "key", {
```
Kurt-von-Laven, line 72: Commit message: the scope can be more specific in this case (e.g., `docker`) since only one file is modified.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM**

PR #129, opened 2022-08-29
docs(integration.test): Move comment for clarity

mwarres:
Move comment next to the precise line of code it describes.


Kurt-von-Laven (changes requested):
I would scope this as `integration.test` for consistency with our existing revision history.

Kurt-von-Laven (approved):
**LGTM**

PR #131, opened 2022-08-31
test(integration): Assert cache not saved on hit

mwarres:
Add assert that `cache.saveCache` is not called upon cache hit.


src/integration.test.ts
```diff
@@ -122,6 +122,7 @@ describe("Integration Test", (): void => {
       `Cache hit occurred on the primary key ${KEY}, not saving cache.`
     );
     expect(child_process.exec).not.toHaveBeenCalled();
+    expect(cache.saveCache).not.toHaveBeenCalled();
```
Kurt-von-Laven, line 125: I'd move it one line down for parity with the production code.

Kurt-von-Laven (changes requested):

PR #134, opened 2022-09-02
test(integration): Add `child_process.exec` assert

mwarres:
Assert `child_process.exec` called precisely once in `mockedExec`.

### DIFF
--- a/src/integration.test.ts
+++ b/src/integration.test.ts
@@ -79,6 +79,7 @@ describe("Integration Test", (): void => {
       EXEC_OPTIONS,
       expect.anything()
     );
+    expect(child_process.exec).toHaveBeenCalledTimes(1);
     expect(core.info).lastCalledWith(stdout);
     expect(core.error).lastCalledWith(stderr);
     expect(core.setFailed).not.toHaveBeenCalled();


src/integration.test.ts
```diff
@@ -82,6 +82,7 @@ describe("Integration Test", (): void => {
     expect(core.info).lastCalledWith(stdout);
     expect(core.error).lastCalledWith(stderr);
     expect(core.setFailed).not.toHaveBeenCalled();
+    expect(child_process.exec).toHaveBeenCalledTimes(1);
```
Kurt-von-Laven, line 85: Nit: I can see the logic of only asserting call counts after all of the called withs have been asserted in order to maximize the helpfulness of the test output, but it also makes it more difficult to trace the correspondence to the production code and correctly update tests when a mock's called with assert is separated from its call count assert. I don't have a strong opinion on which way is objectively preferable overall, so I would just keep consistency with the existing code base for now, meaning co-locate asserts on the same mock. (That seems easier than adopting this new convention throughout the code base after all.)

Kurt-von-Laven (changes requested):
You can use markdown in pull request descriptions, so I formatted the function name as code.

Kurt-von-Laven (changes requested):
You can use markdown in pull request descriptions, so I formatted the function name as code.

Kurt-von-Laven (approved):
**LGTM**

PR #144, opened 2022-09-14
docs(README): Document filtering Docker images

mwarres:
Functionality of docker-cache was changed to filter out images present prior to main step. Document accordingly.


Kurt-von-Laven (changes requested):
Just needs a rebase on the build fix.

Kurt-von-Laven (approved):
**LGTM**

PR #146, opened 2022-09-15
test(action): Use fast-check property-based tests

Kurt-von-Laven:
Property-based tests automatically generate a variety of inputs likely to expose bugs. Replace our conventional tests with property-based tests, and stop hard-coding individual test cases except where needed to ensure we maintain 100% test coverage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@actions/core": "^1.9.1"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


davisgtaylor (approved):
looks good to me

Kurt-von-Laven:
Rebased, corrected the order of the imports in the tests, and upgraded to the latest versions of the new dependencies. Not worth a second look.

PR #149, opened 2022-09-21
ci(ESLint): Enable ESLint by adding config

Kurt-von-Laven:
ESLint as well as its TypeScript and Prettier plugins already ship with the JavaScript flavor of MegaLinter, but ESLint cannot run without being configured. Temporarily work around a MegaLinter issue that prevents ESLint from seeing `tsconfig.json` by configuring MegaLinter to run it in `project` mode.


Kurt-von-Laven:
Please ignore this pull request for now. ESLint isn't actually running even though MegaLinter reports that it is. I will try again after upgrading to the latest version of MegaLinter, but there is likely some additional debugging required here.

mwarres (approved):
LGTM!

PR #156, opened 2022-09-28
chore(Yarn): Install deps for MacOS and current OS

Kurt-von-Laven:
Jest depends on fsevents, which is MacOS-specific. Hence, Yarn doesn't install it on other operating systems by default. This prevents us from running Pivotal Labs' LicenseFinder on this repository, which scans the licenses of the locally installed plugins. Force Yarn to install MacOS-dependencies so they can't break license auditing in the future.


mwarres (approved):
LGTM!

Kurt-von-Laven:
Corrected a typo in the commit message; not worth a second look.

PR #157, opened 2022-09-29
chore(Dependabot): Add Yarn

Kurt-von-Laven:
Automatically update Yarn dependencies via Dependabot. Dependabot recently enabled Yarn v2+ support in production behind a feature flag. We have the privilege of being part of the first batch of external repos to receive access! Use the `fix` prefix in commit messages for bumps to production dependencies so Commitizen bumps the patch version. Use the `chore` prefix in commit messages for bumps to development dependencies so that Commitizen doesn't bump the project version.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -15,6 +15,42 @@ updates:
       prefix: chore
       include: scope
 
+  # Upgrade Yarn production dependencies.
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: daily
+      time: "08:00"
+      timezone: America/New_York
+    assignees:
+      - Kurt-von-Laven
+    reviewers:
+      - Kurt-von-Laven
+    open-pull-requests-limit: 1
+    allow:
+      - dependency-type: production
+    commit-message:
+      prefix: fix
+      include: scope
+
+  # Upgrade Yarn dev dependencies.
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: daily
+      time: "08:00"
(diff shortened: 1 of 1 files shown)


mwarres (approved):
LGTM!

.github/dependabot.yaml
```diff
@@ -15,6 +15,42 @@ updates:
...
+      - Kurt-von-Laven
+    open-pull-requests-limit: 1
+    allow:
+      - dependency-type: production
+    commit-message:
+      prefix: fix
+      include: scope
+
+  # Upgrade Yarn dev dependencies.
+  - package-ecosystem: npm
```
jurre, line 37: Hey @Kurt-von-Laven, it looks like this config file is invalid, you need a single entry per package-ecosystem/directory/target-branch combo. You can get the desired behavior of having separate prefixes by using the `prefix` and `prefix-development` options for the `commit-message` config as described here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message

Right now this is preventing dependabot from picking up the new config 😬

Kurt-von-Laven: Thanks for the tip! And what timing. I was just sitting down to fix this, but you spared me the research. I applied your advice in #159.

Kurt-von-Laven: Also, this makes me suspect that the automatic Dependabot check only runs when you add a new `.github/dependabot.yaml` and not when you modify an existing one. I can open an issue about that if it isn't already a known issue.

PR #158, opened 2022-09-29
ci(workflow): Cache Yarn dependencies

Kurt-von-Laven:
Improve performance of the Test workflow. Yarn dependencies are already installed via the `yarn-install` pre-commit hook.


davisgtaylor (approved):
No issues noted

PR #159, opened 2022-09-29
chore(Dependabot): Correct Yarn configuration

Kurt-von-Laven:
Dependabot complains: "Update configs must have a unique combination of 'package-ecosystem', 'directory', and 'target-branch'." Hence, combine the configurations for Yarn production and development dependencies.


mwarres (approved):
LGTM! And nice commit message!

PR #161, opened 2022-09-29
chore(MegaLinter): Upgrade from v5.16.1 to v6.10.0

mwarres:
Replace MegaLinter v5 hooks with v6 hooks.
Suppress false positive on newly-added linter, Checkov. Change `secrets.GITHUB_TOKEN` to `github.token` in `test.yaml` to avoid triggering a false positive from Checkov.
Add `.trivyignore` because newly-added linter Trivy warns: "Specify at least 1 USER command in Dockerfile with non-root user as argument"; see [relevant Trivy documentation](https://avd.aquasec.com/misconfig/dockerfile/general/avd-ds-0002/). We only use the `Dockerfile` for testing, thus we are not concerned with this as a security issue.
Remove "deps" from custom dictionary; it was added to the Python dictionary upstream.
Remove workarounds for rhysd/actionlint#170 as this was fixed upstream.

### DIFF
--- a/.dictionary.txt
+++ b/.dictionary.txt
@@ -1,3 +1,3 @@
-deps
+checkov
 Laven
 setuptools
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,8 +42,7 @@ jobs:
   save-cache:
     name: Save Cache
     runs-on: ubuntu-22.04
-    permissions:
-      contents: none # workaround for https://github.com/rhysd/actionlint/issues/170
+    permissions: {}
     steps:
       - name: Check out repository.
         uses: actions/checkout@v3.0.2
@@ -80,7 +79,7 @@ jobs:
           --show-error
           --request DELETE
           --header "Accept: application/vnd.github.v3+json"
-          --header "Authorization: token ${{ secrets.GITHUB_TOKEN }}"
+          --header "Authorization: token ${{ github.token }}"
           "https://api.github.com/repos/ScribeMD/docker-cache/actions/caches?key=docker-cache-test&ref=$GITHUB_REF"
   notify:
     name: Notify
@@ -90,8 +89,7 @@ jobs:
       - save-cache
       - restore-cache
     runs-on: ubuntu-22.04
-    permissions:
-      contents: none # workaround for https://github.com/rhysd/actionlint/issues/170
+    permissions: {}
     steps:
       - name: Send Slack notification with workflow result.
         uses: ScribeMD/slack-templates@0.6.8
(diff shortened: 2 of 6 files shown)


.github/workflows/test.yaml
```diff
@@ -73,6 +72,7 @@ jobs:
           [[ $description == "empty image" ]]
       - name: Delete test cache.
         if: always()
+        # checkov:skip=CKV_GHA_3: Suppress error "Suspicious use of curl with secrets" for Delete test cache step.
```
Kurt-von-Laven, line 75: Moot: Omit "for Delete test cache step" from the comment since this is hopefully self-evident from the placement of the comment.

Kurt-von-Laven (changes requested):

Dockerfile
```diff
@@ -1,4 +1,5 @@
 # for testing purposes
+# checkov:skip=CKV_DOCKER_3:Suppress error "Ensure that a user for the container has been created" since this Dockerfile is only used for testing.
```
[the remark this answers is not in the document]
mwarres: I agree with your style sense.  Alas, I followed [the guidelines in Checkov's docs](https://www.checkov.io/2.Basics/Suppressing%20and%20Skipping%20Policies.html); no space following the colon is the format.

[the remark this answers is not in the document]
mwarres: I don't think Checkov supports multiline comments from what I read on the docs (same link as above), but if you're able to find otherwise, that would certainly look nicer.

.github/workflows/test.yaml
```diff
@@ -80,7 +79,7 @@ jobs:
           --show-error
           --request DELETE
           --header "Accept: application/vnd.github.v3+json"
-          --header "Authorization: token ${{ secrets.GITHUB_TOKEN }}"
+          --header "Authorization: token ${{ github.token }}"
```
Kurt-von-Laven, line 82: I would mention the rationale for this change in the commit message since otherwise it seems random.

.dictionary.txt
```diff
@@ -1,3 +1,3 @@
-deps
```
Kurt-von-Laven, line 1: Commit message: "positives" --> "positive"

Kurt-von-Laven (changes requested):

.github/workflows/test.yaml
Kurt-von-Laven, line 82: Commit message: worth noting specifically that you are avoiding a false positive since "triggering" could refer to either a false or true positive.

Kurt-von-Laven (changes requested):

.github/workflows/test.yaml
Kurt-von-Laven, line 82: Commit message: it's typically assumed that relative filepaths are relative to the root of the repository, so I would either say `.github/workflows/test.yaml` or simply `test.yaml` in this case, since no other file in this repository bears that name.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM**

PR #163, opened 2022-09-30
test(action): Remove unneeded bespoke `Mocked` type

Kurt-von-Laven:
Jest v29 recently shipped with stronger typing. The custom `Mocked` type was never necessary. Replace it with `jest.MockedObject` since we will otherwise encounter errors on upgrade.


mwarres (approved):
LGTM!

PR #173, opened 2022-10-02
ci(ESLint): Ignore config files above project dir

Kurt-von-Laven:
Prevent ESLint from reading any config files developers may have in ancestral directories to ensure consistency between development environments.


mwarres (approved):
LGTM!

PR #177, opened 2022-10-03
ci(dictionary): Remove setuptools from custom dict

mwarres:
It is now in the CSpell Python dictionary upstream.


Kurt-von-Laven (approved):
**LGTM**

PR #187, opened 2022-10-06
chore(yarnrc): Add blank line to separate sections

Kurt-von-Laven:
This reduces diff noise when running `yarn set version stable`.


mwarres (approved):
LGTM!

PR #194, opened 2022-10-07
ci(workflow): Don't hardcode GitHub API URL

Kurt-von-Laven:
Prefer the environment variable `$GITHUB_API_URL` for the sake of succinctness.


mwarres (approved):
LGTM!

PR #195, opened 2022-10-07
ci(workflow): Prefer "Bearer" to "token" for auth

Kurt-von-Laven:
"Bearer" is recommended by the official GitHub documentation, because it works with JSON web tokens (JWT).


mwarres (approved):
LGTM!

PR #196, opened 2022-10-07
ci(workflow): Prefer `'` to `"` for string literals

Kurt-von-Laven:
Single quotes prevent evaluation of Bash syntax, such as variable interpolation, so they are simpler and safer when variable interpolation is not desired.


mwarres (approved):
LGTM!

PR #222, opened 2022-10-18
chore(MegaLinter): Sort non-tool-specific keys

Kurt-von-Laven:
Keep config keys that pertain to all tools in alphabetical order.


mwarres (approved):
LGTM!

PR #223, opened 2022-10-18
ci(MegaLinter): Clear report before each run

Kurt-von-Laven:
Preclude confusion regarding which logs were generated by which run.


mwarres (approved):
LGTM!

PR #244, opened 2022-10-27
ci(workflow): Print error messages from GitHub API

Kurt-von-Laven:
Replace `--fail` with `--fail-with-body` in cURL invocation that deletes test cache. This causes cURL to display any error message it may receive from the server, which is often helpful context to have when debugging.


mwarres (approved):
LGTM!

PR #245, opened 2022-10-27
ci(workflow): Support running workflow from forks

Kurt-von-Laven:
The Restore Cache job fails when run from a forked repository, because forks have read-only permissions to GitHub APIs for security purposes. Suppress the failure in the failing step that deletes the test cache since the test cache is less than 2 KB in size. Make the test cache key unique to each workflow run attempt in order to continue testing cache saves after a run from a forked repository. This also eliminates a short race condition that previously could have resulted in false positives when the workflow was run concurrently with itself.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -63,7 +63,7 @@ jobs:
       - name: Save Docker images.
         uses: ./
         with:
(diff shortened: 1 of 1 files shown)


mwarres (approved):
LGTM!

PR #246, opened 2022-10-27
ci(workflow): Add run ID to test cache key

Kurt-von-Laven:
Unlike the run attempt count, the run ID is unique across workflow runs. In fact, unless a workflow is re-run, the run attempt count is always 1. Keep the run attempt count in the key to avoid collisions when the workflow is re-run.


mwarres (approved):
LGTM!

PR #256, opened 2022-11-02
ci(workflow): Make Docker cache key more precise

Kurt-von-Laven:
Only invalidate the Docker cache when the MegaLinter flavor or release change. Previously the cache was invalidated any time `.pre-commit-config.yaml` was modified. This made the cache less efficient since the pre-commit config is frequently modified.


mwarres (approved):
Good change! LGTM!

PR #266, opened 2022-11-07
chore(MegaLinter): Simplify exclude regex

Kurt-von-Laven:
Use the same pattern as the pre-commit exclude regex for consistency.


mwarres (approved):
LGTM!

PR #274, opened 2022-11-08
chore(Commitizen): Don't bump major version

Kurt-von-Laven:
Normally, Commitizen, in compliance with semantic versioning, bumps the project's major version on breaking changes. At major version 0, any release can have breaking changes, and bumping the major version to 1 signifies the stability of the public API. We are currently at major version 0, so configure Commitizen not to bump the major version on breaking changes.


mwarres (approved):
LGTM!

PR #296, opened 2022-11-26
ci(pre-commit): Forbid all Git submodules

Kurt-von-Laven:
Switch from forbid-new-submodules to the newly added forbid-submodules pre-commit hook.


mwarres (approved):
LGTM!

PR #298, opened 2022-11-28
chore(Yarn): Pin all dependencies by removing `^`/`~`

Kurt-von-Laven:
Using `^` or `~` allows the dependency version in use to exceed the version listed in `package.json`. When Renovate performs lock file maintenance, yarn.lock is updated with the latest version of both direct and transitive dependencies, but `package.json` is not updated. The extra complexity of permitting `package.json` and `yarn.lock` to differ provides no value since we use Renovate to keep all of our dependencies up to date. Furthermore, using `^` or `~` can lead to direct dependency bumps being included in Renovate lock file maintenance rather than a dedicated pull request. Renovate subsequently opens pull requests that update `package.json` accordingly but with confusing commit messages based on the lock file stating the direct dependency was bumped from `x.y.z` to `x.y.z`. Pinning dependencies keeps `package.json` and `yarn.lock` in sync, and ensures that Renovate updates both in tandem in dedicated pull requests for direct dependencies.


mwarres (approved):
LGTM!

mwarres (approved):
LGTM!

PR #300, opened 2022-11-28
chore(Poetry): Use `==` to specify exact versions

Kurt-von-Laven:
While Poetry appears to interpret the missing version constraints as exact version requirements, the official documentation requires use of a version constraint, such as `==`.


mwarres (approved):
LGTM!

PR #306, opened 2023-01-07
Problem: macOS doesn't have /usr/bin/bash

yrashk:
Solution: Specify no shell, default to /bin/sh


src/util.ts
```diff
@@ -8,7 +8,7 @@ const execBashCommand = async (command: string): Promise<string> => {
   info(command);
   let output = "";
   try {
-    const result = await execAsPromised(command, { shell: "/usr/bin/bash" });
+    const result = await execAsPromised(command);
```
Kurt-von-Laven, line 11: According to [the Node.js docs](https://nodejs.org/docs/latest-v16.x/api/child_process.html#default-windows-shell):

> Although Microsoft specifies `%COMSPEC%` must contain the path to `'cmd.exe'` in the root environment, child processes are not always subject to the same requirement. Thus, in `child_process` functions where a shell can be spawned, `'cmd.exe'` is used as a fallback if `process.env.ComSpec` is unavailable.

So, I believe this change either eliminates or at least reduces Windows support. I wonder if `"/bin/sh"` or similar might work on all platforms. The most expedient way I can think of to test alternate shell paths would be to use [a matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy) to run the [Save Cache](https://github.com/ScribeMD/docker-cache/blob/c660dd11c0a6c8244d41c2932746f8bbb27aa188/.github/workflows/test.yaml#L77) and [Restore Cache](https://github.com/ScribeMD/docker-cache/blob/c660dd11c0a6c8244d41c2932746f8bbb27aa188/.github/workflows/test.yaml#L92) jobs on [ubuntu-22.04, macos-12, and windows-2022](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). This would necessitate replacing `docker-cache-test-${{ github.run_id }}-${{ github.run_attempt }}` with `docker-cache-test-${{ matrix.os }}-${{ github.run_id }}-${{ github.run_attempt }}` to ensure no unintended collisions in the cache key used by the [Save Cache](https://github.com/ScribeMD/docker-cache/blob/c660dd11c0a6c8244d41c2932746f8bbb27aa188/.github/workflows/test.yaml#L85) and [Restore Cache](https://github.com/ScribeMD/docker-cache/blob/c660dd11c0a6c8244d41c2932746f8bbb27aa188/.github/workflows/test.yaml#L101) steps.

Kurt-von-Laven (changes requested):
Great idea; neat to see someone trying to run this action on MacOS for the first time. Please take a look at [the contributing guide](https://github.com/ScribeMD/.github/blob/75505b6ce742474da60d4f2fccd0d42aa6a18839/CONTRIBUTING.md), and feel free to let us know if you have any feedback or encounter any issues. This change implies an update to the ["Supported Runners" section of the README](https://github.com/ScribeMD/docker-cache/blob/1558650ad4e66e4f778a57a035013e50ab5c1dd7/README.md#supported-runners).

src/util.ts
yrashk: I am not sure I understand the docs in that part very well (as in, how the removal of `shell: "/usr/bin/bash"` affects Windows operations).

Kurt-von-Laven: I take it to mean that Windows typically defaults to a non-Bash shell, such as PowerShell or Command Prompt.

yrashk: If you specify /usr/bin/bash, that wouldn't work on Windows anyway, would it? (I am unfortunately not very familiar with intricacies of Windows toolchains)

Kurt-von-Laven: Yes, it does; that's actually why we bothered to specify the shell in the first place. Windows ships with Bash these days.

yrashk: Does it ship with /bin/sh?

Kurt-von-Laven: Yes, as well as `/bin/bash`, which would be slightly preferable if it works on MacOS so that Bash syntax is supported. I should clarify that when I said Windows, I was and am really referring to Windows Subsystem for Linux (WSL), which is why the Unix-style paths are supported. Windows itself does also ship with Bash, but it isn’t accessible via a Unix-style path without WSL or some other interface. That being said, [the `windows-2022` image does currently include WSLv1](https://github.com/actions/runner-images/blob/68f38ef5a48c784520a1f44819282a7ff938f86f/images/win/Windows2022-Readme.md#windows-features), so I’m comfortable requiring WSL as a dependency on Windows.

Kurt-von-Laven: Unfortunately, [the `macos-12` image doesn't ship with Docker](https://github.com/actions/runner-images/blob/751fe08d9840d2273fb1986980c5f18f3a920e64/images/macos/macos-12-Readme.md), so docker-cache can't be used on macOS. @mwarres is adding Windows support in #347 though.

Kurt-von-Laven: Windows support has been released in 0.3.0. I am closing this pull request as untenable for now. Please feel free to reach back out if GitHub Actions adds Docker to their macOS runners or you find some other way to achieve macOS support.

Kurt-von-Laven: Unfortunately [there is a licensing issue preventing GitHub Actions from adding Docker to their macOS runners](https://github.com/actions/runner-images/issues/17#issuecomment-614726536).

yrashk: Can this be merged? Self-hosted runners are not affected by the above issue.

Kurt-von-Laven: Not as is:

> I believe this change either eliminates or at least reduces Windows support.

I am happy to accept a pull request that adds support for self-hosted macOS runners in a reasonably general way without eliminating Windows support.

PR #347, opened 2023-02-01
feat(OS): Add Windows support

mwarres:


src/integration.test.ts
```diff
@@ -16,7 +16,7 @@ const getKey = (paths: string[], key: string): string => {
 };
 
 describe("Integration Test", (): void => {
-  const EXEC_OPTIONS = { shell: "/usr/bin/bash" };
+  const EXEC_OPTIONS = { shell: "/bin/sh" };
```
Kurt-von-Laven, line 19: I can see that [macos-12 ships with Bash](https://github.com/actions/runner-images/blob/17295fdcdeefbcac13560031c8ee222d84085bd3/images/macos/macos-12-Readme.md). Maybe try running `whereis bash` and/or `which bash` to ask for the appropriate path on each OS. You can do this by adding a step to the Test workflow before the point at which it's crashing.

Kurt-von-Laven (changes requested):

.github/workflows/test.yaml
```diff
@@ -74,22 +74,31 @@ jobs:
           comment_mode: off
   save-cache:
     name: Save Cache
-    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, windows-2022]
```
Kurt-von-Laven, line 79: Nit (multiple instances): prefer the following style for consistency:

```yaml
os:
  - ubuntu-22.04
  - windows-2022
```

The principal benefits of this style are that it makes diffs that modify the elements of a list more readable and also makes it simpler to write regexes that match a specific line (e.g., to configure tools like Renovate).

src/integration.test.ts
```diff
@@ -16,7 +16,7 @@ const getKey = (paths: string[], key: string): string => {
 };
 
 describe("Integration Test", (): void => {
-  const EXEC_OPTIONS = { shell: "/usr/bin/bash" };
+  const EXEC_OPTIONS = { shell: "/bin/bash" };
```
Kurt-von-Laven, line 19: I wonder what happens if we stick with `/usr/bin/bash` on Linux but use `C:\Program Files\Git\bin\bash.exe` on Windows. That path to Git Bash is what the GitHub Actions runner uses when `shell: bash` is specified. Keep in mind that you will have to escape the backslashes.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (changes requested):
The checks didn't run because this branch needs a rebase.

Kurt-von-Laven:
The test failures were caused by the fact that there are Git rebase conflict tokens in `dist/`. This can be addressed by running `yarn build`.

src/integration.test.ts
mwarres: It fails the integration test; I get the following output from the integration test:
```
Stack trace: Error: expect(jest.fn()).nthCalledWith(n, ...expected)

    n: 1
    - Expected
    + Received

      "docker image list --format \"{{ .Repository }}:{{ .Tag }}\"",
      Object {
    -   "shell": "C:\\Program Files\\Git\\bin\\bash.exe",
    +   "shell": "/bin/bash",
      },
      [Function anonymous],

    Number of calls: 1

      101 |   ): void => {
      102 |     expect(core.info).nthCalledWith<[string]>(infoCallNum, command);
    > 103 |     expect(child_process.exec).nthCalledWith<[string, ExecOptions, any]>(
          |                                ^
      104 |       execCallNum,
      105 |       command,
      106 |       EXEC_OPTIONS,

      at assertExecBashCommand (integration.test.ts:103:32)
      at assertLoadDockerImages (integration.test.ts:133:7)
```

.github/workflows/test.yaml
mwarres: I tried this syntax: 
```
runs-on:
      os:
        - ubuntu-22.04
        - windows-2022
```
and got the following error: 
 "runs-on" section must be sequence node but got mapping node with "!!map" tag [syntax-check]
   |
79 |       os:

Is there a different way I should try it?

src/integration.test.ts
mwarres: Now I am trying to do branching on Linux and Windows.  When I run the integration test via `yarn test integration.test.ts`, I get:
```
 Stack trace: Error: expect(jest.fn()).nthCalledWith(n, ...expected)

    n: 1
    - Expected
    + Received

      "docker image list --format \"{{ .Repository }}:{{ .Tag }}\"",
      Object {
    -   "shell": "/usr/bin/bash",
    +   "shell": "/bin/bash",
      },
      [Function anonymous],

    Number of calls: 1

      106 |   ): void => {
      107 |     expect(core.info).nthCalledWith<[string]>(infoCallNum, command);
    > 108 |     expect(child_process.exec).nthCalledWith<[string, ExecOptions, any]>(
          |                                ^
      109 |       execCallNum,
      110 |       command,
      111 |       EXEC_OPTIONS,

      at assertExecBashCommand (integration.test.ts:108:32)
      at assertLoadDockerImages (integration.test.ts:138:7)
      at integration.test.ts:216:7
```
I used the following ternary in my attempt to branch on the os:
```
const EXEC_OPTIONS =
    process.platform === "linux"
      ? { shell: "/usr/bin/bash" }
      : { shell: "/bin/bash" };
```
I used console.logs/Console Ninja to confirm that the integration test is running on "linux" and that `EXEC_OPTIONS` is `"/usr/bin/bash"`.  I am a bit befuddled because I thought "/usr/bin/bash" was what it was before we tried to add in MacOS/Windows support and that seemed to work then.  What am I missing?

Kurt-von-Laven: The idea was to try using `C:\\Program Files\\Git\\bin\\bash.exe` on Windows since we already saw that `/bin/bash` didn't work. Are you setting `EXEC_OPTIONS` via branching in the test or the production code? We don't need the tests to pass though just to check whether that's the correct path to Bash in the windows-2022 GitHub Actions runner image.

.github/workflows/test.yaml
Kurt-von-Laven: [You are missing the `strategy` and `matrix` keys and should remove `runs-on`](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs).

mwarres: I tried removing `runs-on` and VS Code was unhappy with me--it had an error saying `Missing property "runs-on".yaml-schema: GitHub Workflow`.  Just to make sure, I ran it through MegaLinter, which also agreed.  When I try it with the following syntax:
```
strategy:
      matrix:
        os:
          - ubuntu-22.04
          - windows-2022
 runs-on: ${{ matrix.os }}
```
both VS Code and MegaLinter seem to be happy. I got that syntax [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-multi-dimension-matrix). It says it's for a multi-dimension matrix but it seems to apply to me.  Do you think it's ok?

mwarres:
The `Save Cache` job fails on the `Build an empty test docker image` step with the error:
```
Run docker build --tag empty .
Sending build context to Docker daemon  [8](https://github.com/ScribeMD/docker-cache/actions/runs/4255377872/jobs/7402908057#step:5:9).988MB

Step 1/3 : FROM scratch
Windows does not support FROM scratch
Error: Process completed with exit code 1.
```
I'm not sure how to proceed.  This [stack overflow](https://stackoverflow.com/questions/58018675/how-to-create-a-dockerfile-from-scratch-for-windows-exe-file-which-does-not-hav) seems relevant but I definitely don't know how to translate this into anything actionable with the code.  Thoughts?

Kurt-von-Laven:
Yay! That means it worked! Now, it's simply a matter of addressing the remaining failures. The first thing I would try is replacing `scratch` with [`hello-world`](https://hub.docker.com/_/hello-world).

src/integration.test.ts
```diff
@@ -16,7 +16,10 @@ const getKey = (paths: string[], key: string): string => {
 };
 
 describe("Integration Test", (): void => {
-  const EXEC_OPTIONS = { shell: "/usr/bin/bash" };
+  const execOptions =
+    process.platform === "linux"
+      ? { shell: "/usr/bin/bash" }
+      : { shell: "C:\\Program Files\\Git\\bin\\bash.exe" };
```
Kurt-von-Laven, line 22: Nit (multiple instances): I believe you can branch within the object rather than outside it to reduce duplication. Also, the Windows path definitely will never work on any other platform, so I would compare `process.platform` against `"win32"` rather than `"linux"`. The code as written would break macOS development environments for example.

.github/workflows/test.yaml
```diff
@@ -75,22 +75,35 @@ jobs:
...
+      matrix:
+        os:
+          - ubuntu-22.04
+          - windows-2022
+    runs-on: ${{ matrix.os }}
     permissions: {}
     steps:
+      - name: Find Bash path.
+        run: which bash
+        shell: bash
```
Kurt-von-Laven, line 88: We can get rid of this now that we found the correct path on Windows.

Kurt-von-Laven (changes requested):

mwarres:
I initially tried `FROM hello-world`, but Hadolint indicated that I needed a specific tag for the Docker image. I tried `FROM hello-world:linux` which, predictably, worked on Linux but not Windows. Next, I tried `FROM hello-world:nanoserver-ltsc2022`, which, also predictably, works on Windows but not Linux. I've been googling how to do branching logic in Dockerfiles in hopes that I can branch on `process.platform` but I thus far haven't figured out anything viable.  What do you think?

src/util.test.ts
```diff
@@ -38,7 +38,12 @@ describe("Util", (): void => {
       expect(core.info).nthCalledWith<[string]>(1, command);
       expect(child_process.exec).lastCalledWith(
         command,
-        { shell: "/usr/bin/bash" },
+        {
+          shell:
+            process.platform === "win32"
+              ? "C:\\Program Files\\Git\\bin\\bash.exe"
+              : "/usr/bin/bash",
+        },
```
Kurt-von-Laven, line 46: Nit (multiple instances): it's probably worth having this defined in its own `execOptions` variable for readability. The less the brain has to absorb per line, the easier (within reason).

Kurt-von-Laven (changes requested):
Great discovery! Rather than branch within the Dockerfile, which is made painful by the absence of Docker buildx on the Windows runners, I would create a new `Dockerfile.windows`. We might as well go back to building from scratch on Linux, and mcr.microsoft.com/windows/nanoserver:ltsc2022 seems likely to be the optimal image to build from on Windows since it's presumably the smallest of the [pre-cached Docker images](https://github.com/actions/runner-images/blob/bcb2886e5b0fb66c491f66640461128755c06885/images/win/Windows2022-Readme.md#cached-docker-images).

src/util.test.ts
mwarres: I made a `shellPath` variable, which I had done in some of the code, to simplify it.  I hope that accomplishes what you are hoping for.  Let me know. It seemed like `execOptions` would end up being pretty wordy if I didn't introduce an extra variable in some parts of the code.

mwarres:
If I have two files, `Dockerfile` and `Dockerfile.windows`, will test workflow automatically use `Dockerfile` for Linux and `Dockerfile.windows` for Windows? Or do I need to incorporate some branching logic somewhere.  My googling hasn't elucidated anything. I think I'll try it and see if it works.

src/integration.test.ts
```diff
@@ -16,7 +16,11 @@ const getKey = (paths: string[], key: string): string => {
 };
 
 describe("Integration Test", (): void => {
-  const EXEC_OPTIONS = { shell: "/usr/bin/bash" };
+  const shellPath =
+    process.platform === "win32"
+      ? "C:\\Program Files\\Git\\bin\\bash.exe"
+      : "/usr/bin/bash";
```
Kurt-von-Laven, line 22: Multiple instances: the Jest coverage report complains about these branches in the test files since the coverage report is generated on Linux. In `util.test.ts`, I suggest mocking `process.platform` so that both the Windows and non-Windows flows are exercised in testing. In `integration.test.ts`, it would be sufficient to revert the changes since the existing test should pass regardless of what platform it was run on.

Kurt-von-Laven (changes requested):
Unfortunately, no. You would need to run `docker build --tag empty .` or `docker build --tag empty . --file Dockerfile` on Linux and `docker build --tag empty . --file Dockerfile.windows` on Windows. This can be achieved by setting a variable in the same step as the `docker build` command itself based on the value of `${{ matrix.os }}`.

mwarres:
Here is a tentative solution to the Bash logic that involves an intermediate variable as you suggested.
```
run: |
          if [ "${{ matrix.os }}" = "ubuntu-22.04" ]; then
            DOCKERFILE="Dockerfile"
          elif [ "${{ matrix.os }}" = "windows-2022" ]
            DOCKERFILE="Dockerfile.windows"
          docker build --tag . --file $DOCKERFILE
```
I'm wondering if it is better to write it without the intermediate variable.  I'm thinking something like this:
```
run: |
          if [ "${{ matrix.os }}" = "ubuntu-22.04" ]; then
            docker build --tag . --file Dockerfile
          elif [ "${{ matrix.os }}" = "windows-2022" ]; then
            docker build --tag . --file Dockerfile.windows
```
What do you think?
I'm also wondering if there is appropriate logic that I should incorporate for if the value of `${{ matrix.os }}` is neither the appropriate Linux or Windows OS. I think that currently if the OS didn't match what was expected, the `docker build` command would fail because $DOCKERFILE would be undefined and GitHub Actions would indicate failure. Is this sufficient or should I be incorporating more explicit error handling?

Kurt-von-Laven:
I do prefer the DRYer way even though it adds a line of code. We should probably follow the precedent of the `process.platform` ternaries and handle Windows specially but everything else the same way for now. You can use the `=~` operator instead so you don't have to specify the OS version.

Kurt-von-Laven (changes requested):

mwarres:
I'm confused about the GitHub Actions failure.  I'm seeing this error:
```
if [[ "windows-2022" =~ "windows" ]]; then
     |    ~
     | Missing '(' after 'if' in if statement.
```
I don't see parentheses associated with `if` statements in any of the online examples of conditional logic in Bash.  What do you make of this?

Kurt-von-Laven (changes requested):

Dockerfile.windows
```diff
@@ -0,0 +1,6 @@
+# for testing purposes
+# checkov:skip=CKV_DOCKER_3:Suppress error "Ensure that a user for the container has been created" since this Dockerfile is only used for testing.
+
+FROM mcr.microsoft.com/windows/nanoserver:ltsc2022
+LABEL description="empty image"
+COPY LICENSE /LICENSE
```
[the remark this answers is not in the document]
mwarres: Good catch--I'll try it without the` COPY` statement.

.github/workflows/test.yaml
```diff
@@ -73,22 +73,38 @@ jobs:
...
       - name: Check out repository.
         uses: actions/checkout@v3.3.0
       - name: Save Docker images.
         uses: ./
         with:
-          key: docker-cache-test-${{ github.run_id }}-${{ github.run_attempt }}
+          key: docker-cache-test-${{ matrix.os }}-${{ github.run_id }}-${{ github.run_attempt }}
       - name: Build an empty test Docker image.
-        run: docker build --tag empty .
+        run: |
```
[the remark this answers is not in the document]
mwarres: Thanks--that makes a lot of sense!

src/util.test.ts
```diff
@@ -33,27 +34,36 @@ describe("Util", (): void => {
...
       expect(child_process.exec).lastCalledWith(
         command,
-        { shell: "/usr/bin/bash" },
+        { shell },
         expect.anything()
       );
       return output;
     };
 
+    const mockProcessPlatform = () =>
```
Kurt-von-Laven, line 53: This is a misnomer since this is an arbitrary rather than a mock. In keeping with the naming convention of fast-check, it could be named for what it generates (e.g., `platform()`). To avoid confusion, it would also be good to rename the `os` variables to `platform`.

.github/workflows/test.yaml
```diff
@@ -97,7 +114,7 @@ jobs:
       - name: Load Docker images.
         uses: ./
         with:
-          key: docker-cache-test-${{ github.run_id }}-${{ github.run_attempt }}
+          key: docker-cache-test-${{ matrix.os }}-${{github.run_id }}-${{ github.run_attempt }}
       - name: Verify Docker loaded empty image from cache.
         run: |
```
Kurt-von-Laven, line 119: `shell: bash` is needed in this step and the one below it as well.

Kurt-von-Laven (changes requested):

.dictionary.txt
```diff
@@ -1,3 +1,5 @@
 checkov
 Laven
+ltsc
+nanoserver
```
Kurt-von-Laven, line 4: I added these to the win32 dictionary.

Kurt-von-Laven (changes requested):

.dictionary.txt
mwarres: I tried removing those words and adding the win32 dictionary, but it looks like those words didn't make it into this release.  Hopefully soon!

.github/workflows/test.yaml
```diff
@@ -74,21 +74,38 @@ jobs:
...
         uses: actions/checkout@v3.3.0
       - name: Save Docker images.
         uses: ./
         with:
-          key: docker-cache-test-${{ github.run_id }}-${{ github.run_attempt }}
+          key: docker-cache-test-${{ matrix.os }}-${{ github.run_id }}-${{ github.run_attempt }}
       - name: Build an empty test Docker image.
-        run: docker build --tag empty .
+        run: |
+          if [[ "${{ matrix.os }}" =~ "windows" ]]; then
```
Kurt-von-Laven, line 92: `"windows"` should also be single-quoted since there's no interpolation desired.

src/util.test.ts
```diff
@@ -33,27 +34,41 @@ describe("Util", (): void => {
...
       expect(child_process.exec).lastCalledWith(
         command,
-        { shell: "/usr/bin/bash" },
+        { shell },
         expect.anything()
       );
       return output;
     };
 
+    const platform = () => constantFrom<NodeJS.Platform>("linux", "win32");
```
Kurt-von-Laven, line 53: Please specify the return type since ESLint will likely complain about that once it awakens from its slumber (c.f., oxsecurity/megalinter#1572).

src/util.test.ts
```diff
@@ -33,27 +34,41 @@ describe("Util", (): void => {
       ): any => {
         callback(error, { stdout, stderr });
       }));
-      const output = await util.execBashCommand(command);
+
```
Kurt-von-Laven, line 37: Nit: undo blank line add.

src/util.ts
```diff
@@ -4,11 +4,19 @@ const execAsPromised = promisify(exec);
...
 
-const execBashCommand = async (command: string): Promise<string> => {
+const execBashCommand = async (
+  command: string,
+  platform: NodeJS.Platform = process.platform
+): Promise<string> => {
+  const shell =
+    platform === "win32"
+      ? "C:\\Program Files\\Git\\bin\\bash.exe"
+      : "/usr/bin/bash";
```
Kurt-von-Laven, line 14: Nit: Recommend moving this after `info(command);` since it defines a variable used in the `try` block.

src/util.test.ts
```diff
@@ -24,7 +24,8 @@ describe("Util", (): void => {
       command: string,
       error: Error | null,
       stdout: string = "",
-      stderr: string = ""
+      stderr: string = "",
+      platform: NodeJS.Platform
```
Kurt-von-Laven, line 28: There's no sense in having default arguments followed by an argument without a default in JavaScript since there's no way to use any of the defaults in that case. The default values can be removed if they aren't needed, and the order of the arguments can be switched if they are.

Dockerfile.windows
```diff
@@ -0,0 +1,5 @@
+# for testing purposes
+# checkov:skip=CKV_DOCKER_3:Suppress error "Ensure that a user for the container has been created" since this Dockerfile is only used for testing.
```
Kurt-von-Laven, line 2: Can we configure Checkov to ignore this error in all Dockerfiles so that we don't have to duplicate the suppression?

Kurt-von-Laven (changes requested):

src/util.test.ts
```diff
@@ -33,27 +34,41 @@ describe("Util", (): void => {
...
       );
       return output;
     };
 
+    const platform = () => constantFrom<NodeJS.Platform>("linux", "win32");
+
     testProp(
       "ferries command output to GitHub Actions on success",
-      [string(), string(), string()],
+      [string(), string(), string(), platform()],
```
Kurt-von-Laven, line 57: Multiple instances: we should provide at least two explicit test cases like we do [here](https://github.com/ScribeMD/docker-cache/blob/3c74f60e0b0e1ec4c30fed29f7627e58e714bd0d/src/docker.test.ts#L271-L278) in order to ensure that we don't fail test coverage if `constantFrom` happens to pull the same element every time.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM**

.github/workflows/test.yaml
```diff
@@ -97,11 +114,12 @@ jobs:
...
-          key: docker-cache-test-${{ github.run_id }}-${{ github.run_attempt }}
+          key: docker-cache-test-${{ matrix.os }}-${{github.run_id }}-${{ github.run_attempt }}
       - name: Verify Docker loaded empty image from cache.
         run: |
           description="$(docker inspect --format '{{ index .Config.Labels "description" }}' empty)"
           [[ $description == 'empty image' ]]
+        shell: bash
       - name: Delete test cache if permitted (i.e., workflow not triggered from fork).
         if: always()
         run: >
```
Kurt-von-Laven, line 125: I overlooked the fact that we forgot to update the key in the delete cache step.

mwarres: Oooh thank you for catching this! I'll try to fix it now!

PR #349, opened 2023-02-02
chore(http-cache-semantics): Bump v4.1.0 to v4.1.1

Kurt-von-Laven:
A moderate severity security vulnerability was discovered in http-cache-semantics, a transitive Yarn dependency.


mwarres (approved):
LGTM!

PR #372, opened 2023-02-15
docs(README): Clarify how to create a cache key

### LINKED ISSUE #97: Clarify hashFiles(...) on the README
The README has a key example containing `hashFiles(...)` - what should that path be to capture the docker images? Where do they go by default?

Kurt-von-Laven:
Resolves #97.

Replace `hashFiles(...)` with `hashFiles(paths)` to make it clear that `...` isn't a special symbol and that the user ought to pass one or more filepaths to `hashFiles`. Explicitly document that the user must select an appropriate key and cannot copy the example verbatim. Link to the specific section of the official GitHub cache action that documents how to create a cache key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,14 @@ performs Docker layer caching for built images but does not cache pulled images.
   - name: Cache Docker images.
     uses: ScribeMD/docker-cache@0.2.8
     with:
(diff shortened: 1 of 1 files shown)


marc-hb (approved):
Perfect

mwarres (approved):
LGTM! The example seems like helpful clarification!

PR #377, opened 2023-02-17
ci(workflow): Partially restore Yarn cache

Kurt-von-Laven:
Implement partial cache restoration for Yarn, allowing cache hits when Yarn dependencies are upgraded. Yarn prevents partial cache restoration snowballing by removing no longer needed dependencies from the cache when `yarn install` is run.


mwarres (approved):
LGTM!

PR #387, opened 2023-02-24
chore(tsconfig): Upgrade from node16 to node18

Kurt-von-Laven:
We already upgraded from Node.js 16 to 18, so reflect that in our TypeScript configuration by upgrading from @tsconfig/node16-strictest-esm to @tsconfig/node18-strictest-esm.


mwarres (approved):
LGTM!

PR #388, opened 2023-02-24
test(Jest): Remove `testEnvironment` and `watchman`

Kurt-von-Laven:
`testEnvironment` began defaulting to `"node"` in Jest 27, and we are currently on Jest 29. `watchman` defaulted to `true` all long.


mwarres (commented):
LGTM!

PR #389, opened 2023-02-25
test(jest.config): Stop transforming `*.js` files

Kurt-von-Laven:
Jest only runs on the `*.ts` files in the `src` directory, so there have never been any `*.js` source files for ts-jest to transform. tsc and @vercel/ncc generate `*.js` output files in the `build` and `dist` directories, respectively, but ts-jest doesn't transform these.


mwarres (approved):
LGTM!

PR #390, opened 2023-02-26
ci(workflow): Stop publishing nonexistent report

Kurt-von-Laven:
There was never anything generating a report at a path matching `reports/junit/*.xml`, so instructing EnricoMi/publish-unit-test-result-action to search for one there only served to slow it down.


mwarres (approved):
Good fix! LGTM!

PR #391, opened 2023-02-28
test(Jest): Update paths to reflect `rootDir` routes

Kurt-von-Laven:
`coverageDirectory` is relative to `rootDir`, so Jest outputs test coverage reports to `src/reports/jest/` rather than `reports/jest/`. `outputDirectory` is not relative to `rootDir`, so include `rootDir` explicitly in the path so that unit tests are reported to the same directory as test coverage.


mwarres (approved):
LGTM!

PR #392, opened 2023-02-28
ci(workflow): Replace `junit_files` with `files`

Kurt-von-Laven:
The `junit_files` argument to [EnricoMi/publish-unit-test-result-action](https://github.com/EnricoMi/publish-unit-test-result-action) was deprecated in favor of `files`.


mwarres (approved):
LGTM!

PR #410, opened 2023-03-08
chore(VSCode): Don't trim Git patch whitespace

Kurt-von-Laven:
VSCode and the EditorConfig VSCode extension are configured to trim trailing whitespace. This is generally desirable, but breaks Git patch editing, so disable whitespace trimming for `*.diff` files for both VSCode and EditorConfig.


mwarres (approved):
LGTM!

PR #412, opened 2023-03-08
ci(gitattributes): Ignore `*.mjs` whitespace errors

Kurt-von-Laven:
Yarn generates `.pnp.loader.mjs` for ESM (ECMAScript modules) support. We already correctly configure Git to ignore whitespace errors in Yarn CJS (CommonJS) files, so do the same for `*.mjs` files.


mwarres (approved):
LGTM!

PR #415, opened 2023-03-09
ci(workflow): Set top-level permissions to `{}`

Kurt-von-Laven:
Add minimal permissions to each job as needed rather than start broad and subtract.


mwarres (approved):
LGTM!

PR #420, opened 2023-03-10
chore(VSCode): Set Python to 4-space indentation

Kurt-von-Laven:
Configure VSCode and EditorConfig to follow the PEP 8 style guideline of 4-space indentation in Python. We don't track any Python files in this repository, but Poetry does install Python dependencies to `.venv`.


mwarres (approved):
LGTM!

PR #421, opened 2023-03-10
test(util): Remove unnecessary typing

mwarres:
TypeScript automatically infers the type of `""` to be `string`.


Kurt-von-Laven (approved):
**LGTM**

PR #422, opened 2023-03-11
chore(VSCode): Remove no longer needed exclude

Kurt-von-Laven:
VSCode search now excludes gitignored files by default.


mwarres (approved):
LGTM!

PR #423, opened 2023-03-11
chore(VSCode): Include `poetry.lock` when searching

Kurt-von-Laven:
Occasionally one may desire to search the Poetry lock file. Search results from the lock file may be ignored temporarily by collapsing them or permanently via the user setting `search.exclude`.


mwarres (approved):
LGTM!

PR #424, opened 2023-03-11
test(docker): Remove unnecessary typing

Kurt-von-Laven:
TypeScript automatically infers argument types from default values.


mwarres (approved):
LGTM!

PR #433, opened 2023-03-15
test(action): Remove unnecessary `beforeAll`s

Kurt-von-Laven:
Simplify `main.test.ts` and `post.test.ts` by moving the contents of their `beforeAll`s into their single tests. Use `jest.mocked()` to automatically convert the type of the imported docker modules to `jest.MockedObject<typeof import("./docker.js")>`.


mwarres (approved):
LGTM!

PR #438, opened 2023-03-16
test(action): Prefer `jest.mocked()` to `<any>`

Kurt-von-Laven:
Casting mocked modules to `any` prevents TypeScript from type checking. Jest offers `jest.mocked()` for the purpose of decorating the types of mocked objects with their Jest-imbued functions and properties. This allows TypeScript to enforce that the appropriate mock is assigned to the appropriate variable, for instance.


mwarres (approved):
Good improvement! LGTM!

PR #439, opened 2023-03-16
ci(workflow): Correct cache key for delete step

mwarres:
When we added Windows support, the cache key for the delete step in our test workflow was not updated accordingly.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -131,6 +131,8 @@ jobs:
           --header 'Accept: application/vnd.github.v3+json'
           --header 'Authorization: Bearer ${{ github.token }}'
           "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/actions/caches?key=docker-cache-test-${{
+            matrix.os
+          }}-${{
             github.run_id
           }}-${{
             github.run_attempt


.github/workflows/test.yaml
```diff
@@ -130,7 +130,7 @@ jobs:
           --request DELETE
           --header 'Accept: application/vnd.github.v3+json'
           --header 'Authorization: Bearer ${{ github.token }}'
-          "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/actions/caches?key=docker-cache-test-${{
+          "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/actions/caches?key=docker-cache-test-${{ matrix.os }}-${{
```
Kurt-von-Laven, line 133: Nit: recommend following the existing style of putting each context on its own line since this line is fairly long.

Kurt-von-Laven (changes requested):
Since the Test workflow in this repository isn't callable, we don't want Commitizen to release this change that has no downstream impact. The semantic commit type should be "ci" rather than "fix."

Kurt-von-Laven (approved):
**LGTM** I confirmed that automatic test cache deletion is working once more by manually inspecting the "Checks" tab. That is a good habit for changes to CI, most notably changes in steps where we suppress errors!

Kurt-von-Laven (changes requested):
Just realized the commit scope should be "workflow" rather than "test."

Kurt-von-Laven (approved):
**LGTM**

PR #446, opened 2023-03-19
style(util): Introduce `ConsoleOutput` type

Kurt-von-Laven:
Stdout and stderr frequently co-occur, so wrap them in a dedicated type. Create a corresponding fast-check arbitrary, `consoleOutput()`. Move our existing custom arbitrary, `platform()`, into the new dedicated arbitrary file, `src/arbitraries/util.ts`. Leverage `ConsoleOutput` in a new helper method, `joinOutput()`, to reduce duplication in the integration test.

### DIFF
--- a/.dictionary.txt
+++ b/.dictionary.txt
@@ -1,3 +1,4 @@
+arbitraries
 Laven
 ltsc
 nanoserver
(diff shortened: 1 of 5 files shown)


mwarres (approved):
LGTM!

PR #447, opened 2023-03-19
test(integration): Elide `return` in one-liners

Kurt-von-Laven:
Prefer the shorthand notation for one-line functions in which the `return` statement and curly braces are omitted.


mwarres (approved):
LGTM!

mwarres (approved):
LGTM!

PR #452, opened 2023-03-20
test(integration): Remove incorrect use of `jest.mocked()`

Kurt-von-Laven:
The integration test doesn't mock any of our own production code since that is what it is testing. This error was asymptomatic since we didn't actually mock any of our production code, but now we stop giving the misleading impression that we might.


mwarres (commented):
LGTM!

PR #453, opened 2023-03-20
test(action): Move `async` imports to top level

Kurt-von-Laven:
Top-level `await` allows us to declare and initialize dynamic asynchronous imports at once using `const`. This empowers TypeScript to infer the appropriate types, eliminating the need to specify them explicitly.

### DIFF
--- a/src/docker.test.ts
+++ b/src/docker.test.ts
@@ -12,6 +12,11 @@ jest.unstable_mockModule("./util.js", (): typeof import("./util.js") => ({
   execBashCommand: jest.fn<typeof import("./util.js").execBashCommand>(),
 }));
 
+const cache = jest.mocked(await import("@actions/cache"));
+const core = jest.mocked(await import("@actions/core"));
+const util = jest.mocked(await import("./util.js"));
+const docker = await import("./docker.js");
+
 // Expect the given mocks were called in the given order.
 const assertCalledInOrder = <T extends FunctionLike>(
   ...mocks: jest.MockedFunction<T>[]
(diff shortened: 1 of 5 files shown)


mwarres (approved):
LGTM!

src/docker.test.ts
```diff
@@ -12,6 +12,11 @@ jest.unstable_mockModule("./util.js", (): typeof import("./util.js") => ({
   execBashCommand: jest.fn<typeof import("./util.js").execBashCommand>(),
 }));
 
+const cache = jest.mocked(await import("@actions/cache"));
+const core = jest.mocked(await import("@actions/core"));
+const util = jest.mocked(await import("./util.js"));
```
mwarres, line 17: Question: what is the rationale behind the ordering of the imports here? I'm seeing that in the `describe` block in the previous version of the code, util is before docker.  Then the reverse is true in the `beforeAll` block.  I think we usually do alphabetical order, so I'm wondering if this is a mistake or if there is something I'm missing.

Kurt-von-Laven: Are you comparing different files? In this file, `util` consistently precedes `docker`. We import the unit under test last so that we don't mix it with the mocked dependencies. We import built-in node modules before third-party dependencies before our own dependencies and sort alphabetically within each section.

mwarres: Got it.  That makes sense! I don't know where I was seeing the inconsistency...my brain....hmmmmmmm.  Although I didn't notice the discrepancy between `util` needing to be mocked and `docker` simply needing to be imported.  Thanks for clearing this up!

Kurt-von-Laven: You were probably just comparing different files.

PR #459, opened 2023-03-23
test(action): Import types statically

Kurt-von-Laven:
Prefer static type imports (using the `import` keyword) to dynamic type imports (using the `import()` function). Unlike regular static imports, static type imports do not interfere with mocking as they compile out, so there is no need to delay them until after our mocks are set up. Introduce `types/aliases.ts` to export convenient names for the shapes of our production modules that we import in our tests. We use these types to ensure that our mocks accurately reflect the shapes of their production counterparts.

### DIFF
--- a/src/docker.test.ts
+++ b/src/docker.test.ts
@@ -5,12 +5,16 @@ import { boolean, string, uniqueArray } from "fast-check";
 import type { InputOptions } from "@actions/core";
 import type { FunctionLike } from "jest-mock";
 
+import type { Util } from "../types/aliases.js";
+import type { execBashCommand } from "./util.js";
+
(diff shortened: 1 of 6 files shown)


mwarres (commented):
Looks good as long as test coverage gets resolved.

Kurt-von-Laven:
I fixed the test coverage false positive by moving `src/types/aliases.ts` to `types/aliases.ts` so that Jest wouldn't enforce runtime test coverage of type declarations. Type declarations are tested by running `yarn tsc`, which we already do as part of the `yarn test` command.

mwarres (approved):
LGTM! Good fix on the test coverage!

PR #472, opened 2023-03-28
ci(dictionary): Remove "arbitraries" from custom dictionary

Kurt-von-Laven:
It was added to the software-terms dictionary upstream.


mwarres (approved):
LGTM!

PR #473, opened 2023-03-28
ci(CSpell): Add win32 dictionary

Kurt-von-Laven:
"LTSC" and "Nanoserver" were added to the win32 dictionary upstream. As the win32 dictionary is large, it is not bundled with @cspell/cspell-bundled-dicts, so install it via a MegaLinter pre-command, and instruct CSpell to import it.


mwarres (approved):
LGTM!

PR #475, opened 2023-03-28
ci(CSpell): Add win32 dict as dev dependency

mwarres:
Eliminate the need to manually install the CSpell win32 dict locally in order for the VS Code CSpell extension to have access to that dictionary. Keep the installation pre-command for now; it is a workaround for a bug in MegaLinter where CSpell fails to import the win32 dict when it is run via MegaLinter.


Kurt-von-Laven (changes requested):
The package name is `@cspell/dict-win32`. This is a reasonable change, but it would be even better if we can simply add the dictionary as a dev dependency.

cspell.config.yaml
```diff
@@ -3,8 +3,6 @@ language: en-US
 useGitignore: true
```
Kurt-von-Laven, line 3: We should specify `usePnP: true` since we are using Plug'n'Play, and hopefully that is all that is needed for CSpell to access to our Yarn dependencies.

Kurt-von-Laven (changes requested):
This was really close to what I would've tried. Consider "ci(CSpell): Add win32 dict as dev dependency" for the commit subject. Then we can explain the motivation in the commit body.

cspell.config.yaml
```diff
@@ -1,10 +1,9 @@
 version: "0.2"
 language: en-US
 useGitignore: true
+usePnP: true
 ignorePaths:
   - .vscode
-import:
-  - "@cspell/dict-win32/cspell-ext.json"
```
Kurt-von-Laven, line 7: My bad; I failed to consider that this import would still be needed since `@cspell/dict-win32` isn't bundled with CSpell.

Kurt-von-Laven (changes requested):
Commit message: I would clarify the purpose for which one would have needed to install the win32 dictionary locally since the reader may assume you mean for running MegaLinter locally.

.mega-linter.yaml
```diff
@@ -17,8 +17,6 @@ JSON_PRETTIER_FILE_EXTENSIONS:
 MARKDOWN_MARKDOWNLINT_CONFIG_FILE: .markdownlint.yaml
 REPOSITORY_CHECKOV_CONFIG_FILE: LINTER_DEFAULT
 SPELL_CSPELL_CONFIG_FILE: LINTER_DEFAULT
-SPELL_CSPELL_PRE_COMMANDS:
-  - command: npm install @cspell/dict-win32@2.0.1
```
Kurt-von-Laven, line 21: Given that CSpell is still failing to load the win32 dictionary, I can only assume that it isn't able to access win32 via the Yarn cache when run through MegaLinter for some reason. You can compare the output of `npx cspell@6.31.1 Dockerfile.windows` to establish that CSpell is configured correctly. This might be a MegaLinter bug, but presumably restoring the pre-command will address the CI failure as a workaround.

Kurt-von-Laven (changes requested):
Just realized I've been misspelling "VS Code" everywhere. Doh.

Kurt-von-Laven (changes requested):
Commit message: I would argue that the pre-command is automated rather than manual since MegaLinter reads it from its config file and executes it automatically. I would simply remove the second use of "manual." I would describe the issue explicitly but briefly rather than linking to it. It's relatively common practice to limit use of links in commit messages (or even ban them outright) since they may eventually break, decreasing their value as a historical record.

Kurt-von-Laven (changes requested):
Commit message: I wouldn't say the dictionary is imported via MegaLinter. CSpell is unable to import the dictionary when run via MegaLinter, but MegaLinter correctly makes no attempt to import the dictionary on CSpell's behalf.

Kurt-von-Laven (approved):
**LGTM**

PR #476, opened 2023-03-29
test(integration): Prefer explicit types to `any`

Kurt-von-Laven:
Explicitly type the return type of our mock of `cache.saveCache` and the `value` argument to our mock of `core.saveState`.


mwarres (approved):
LGTM!

PR #480, opened 2023-03-30
fix(util): Don't call `toString()` on `Error`

Kurt-von-Laven:
`core.setFailed()` accepts either a `string` or an `Error` and calls `toString()` on the latter, so simply pass it the `Error`.


mwarres (approved):
LGTM!

PR #481, opened 2023-03-31
style(util): Prefer `unknown` to `any` in catch

Kurt-von-Laven:
Using the `any` type disables the type system, which typescript-eslint objects to, so switch to `unknown`, the only other option permitted by TypeScript.


mwarres (approved):
LGTM!

PR #482, opened 2023-03-31
ci(LicenseFinder): Prevent license violations

Kurt-von-Laven:
Run Pivotal's License Finder as a pre-commit hook. License Finder currently supports Yarn along with numerous other package managers. License policies are configured via `doc/dependency_decisions.yml`. Install the Yarn licenses plugin, which License Finder relies on to list the licenses of dependencies. The License Finder pre-commit hook depends on Ruby, so add Ruby as an asdf tool.


mwarres (approved):
LGTM!

PR #506, opened 2023-04-19
feat(action): Verify no matching cache before save

### LINKED ISSUE #135: Avoid Post Cache-phase when key is already in cache
It would be nice if the post-cache phase checked that the cache key is still missing before it starts the slow operations. This will speed up workflows where runners have different run times (1 min vs 10 min), and in scenarios where some runners have a delayed start (not sure why runner starts are sometimes staggered, might be related to the load on the Github CI).

mwarres:
Fixes #135.

Prior to saving cache, verify that there is no cache with a matching key already cached. Previously, we checked if a cache key matching our cache was present in the load step; if not, we would then proceed to attempt to save the cache in the save step (assuming we were not in read-only mode and that there were indeed new Docker images to save). This was
problematic when the action was run in parallel; in that case, there were multiple, unnecessary cache save attempts when the cache initially missed but was subsequently saved.


mwarres:
Note: this does not have appropriate testing, current status of PR is just production code changes.

src/docker.ts
```diff
@@ -37,24 +40,53 @@ const saveDockerImages = async (): Promise<void> => {
...
+      creationTime?: string;
+      archiveLocation?: string;
+    }
+    interface CacheWithPrivateMethod extends CacheType {
+      getCacheEntry(
+        keys: string[],
+        paths: string[]
+      ): Promise<ArtifactCacheEntry | null>;
+    }
+    const cacheWithPrivateMethod = cache as CacheWithPrivateMethod;
```
Kurt-von-Laven, line 59: Nit: Prefer `<CacheWithPrivateMethod>cache` for consistency.

src/docker.ts
```diff
@@ -37,24 +40,53 @@ const saveDockerImages = async (): Promise<void> => {
...
+     * cache action.
+     */
+    interface ArtifactCacheEntry {
+      cacheKey?: string;
+      scope?: string;
+      cacheVersion?: string;
+      creationTime?: string;
+      archiveLocation?: string;
+    }
+    interface CacheWithPrivateMethod extends CacheType {
```
Kurt-von-Laven, line 53: Nit: prefer a more specific name like `CacheWithGetCacheEntry` since only one private method is currently exposed.

src/docker.ts
```diff
@@ -37,24 +40,53 @@ const saveDockerImages = async (): Promise<void> => {
...
+      getCacheEntry(
+        keys: string[],
+        paths: string[]
+      ): Promise<ArtifactCacheEntry | null>;
+    }
```
Kurt-von-Laven, line 58: Nit: prefer declaring `interfaces` at the top level. They are a static concept handled by the compiler, so it's helpful to declare them statically.

src/docker.ts
```diff
@@ -37,24 +40,53 @@ const saveDockerImages = async (): Promise<void> => {
...
+    const cacheWithPrivateMethod = cache as CacheWithPrivateMethod;
+    const getCacheEntry = cacheWithPrivateMethod["getCacheEntry"];
+    /* Check if a cache with our key has been saved between when we checked in
+     * loadDockerImages and now.
+     */
+    const cacheHit = Boolean(await getCacheEntry([key], [DOCKER_IMAGES_PATH]));
+    if (cacheHit) {
```
Kurt-von-Laven, line 65: The PR can be simplified by combining these lines into a single `else if`, thereby eliminating the indentation changes to the preexisting code.

src/docker.ts
```diff
@@ -37,24 +40,53 @@ const saveDockerImages = async (): Promise<void> => {
...
+      info(
+        "A cache miss occurred in loadDockerImages, but there was a subsequent" +
+          "cache hit when we were in saveDockerImages. Not saving cache."
```
Kurt-von-Laven, line 68: Consider rewording to clarify what this case corresponds to; the reader should comprehend why the cache was already saved. It's slightly misleading to describe this case as a cache hit since we aren't restoring the cache.

types/aliases.ts
```diff
@@ -1,5 +1,7 @@
+import type * as cache from "@actions/cache";
 import type * as docker from "../src/docker.js";
 import type * as util from "../src/util.js";
 
+export type CacheType = typeof cache;
```
Kurt-von-Laven, line 5: Nit: I suppose this alias should be named `Cache` for consistency with the existing ones if that doesn't create any name conflicts.

Kurt-von-Laven (changes requested):
This looks like a great start! I wonder if the failure in the Save Cache job can be addressed by disabling [minification](https://github.com/ScribeMD/docker-cache/blob/374ee92653593d7fae6cc040bf881dde8e2ea7cd/package.json#L13-L14).

Commit message: "was indeed a new docker image(s) to save" --> "were indeed new Docker images to save"

src/docker.ts
mwarres: Sounds good.  Do you want me to change `cacheWithPrivateMethod` to `cacheWithGetCacheEntry` so that I have a more specific name for that as well? Or do you only want me to make the type `CacheWithGetCacheEntry` more specific (per the comment below)?

src/docker.ts
mwarres: Are you wanting me to explicitly mention that this would occur if multiple jobs are run in parallel? Could this also happen if two separate PRs were running through CI using docker cache simultaneously? Are you wanting me to mention these scenarios explicitly? I tentatively changed it to 
```
"A cache miss occurred in loadDockerImages; subsequently a cache with" +
        "a matching key was saved. Not saving cache."
```
but I can mention those specific cases (or others as appropriate) if you'd like me to.

src/docker.ts
Kurt-von-Laven: I would eliminate the intermediate variable `cacheWithPrivateMethod` altogether in order to address the comment about combining these lines into a single `else if`.

src/docker.ts
Kurt-von-Laven: Yeah, I think the parallel job case is worth mentioning. It could also occur with parallel workflows like you described, but I don't think we need to go into that level of detail since folks should generally be able to extrapolate from an example. I like the new wording besides that though.

Kurt-von-Laven (changes requested):
Bummer that ncc replaces `cache["getCacheEntry"]` with `cache.getCacheEntry`. We will have to copy the corresponding code from `@actions/cache` into our repository and modify the visibility of `getCacheEntry` to get this to work.

src/docker.ts
```diff
@@ -36,6 +53,21 @@ const saveDockerImages = async (): Promise<void> => {
...
     );
+    /* Access the private method, getCacheEntry, internal to the GitHub Actions
+     * cache action. Use it to check if a cache with our key has been saved between
+     * when we checked in loadDockerImages and now.
+     */
+  } else if (
+    await (<CacheWithGetCacheEntry>cache)["getCacheEntry"](
+      [key],
+      [DOCKER_IMAGES_PATH]
+    )
```
Kurt-von-Laven, line 64: We could instead test whether [`key === await restoreCache([""], key, [], { lookupOnly: true })`](https://github.com/actions/toolkit/blob/%40actions/cache%403.2.2/packages/cache/src/cache.ts#L56-L72) since [`@actions/cache` added support for `lookupOnly`](https://github.com/actions/toolkit/blob/%40actions/cache%403.2.2/packages/cache/src/options.ts#L63-L70) in [v3.2.0](https://github.com/actions/toolkit/blob/%40actions/cache%403.2.2/packages/cache/RELEASES.md#320). That would let us use the public API of `@actions/cache` rather than jump through hoops to call an internal method and allow us to undo almost all of the other changes.

Kurt-von-Laven (changes requested):

Kurt-von-Laven (changes requested):

src/docker.ts
```diff
@@ -36,6 +36,17 @@ const saveDockerImages = async (): Promise<void> => {
       `Cache miss occurred on the primary key ${key}. Not saving cache as ` +
         "read-only option was selected.",
     );
+    /* Check if a cache with our key has been saved between when we checked in
+     * loadDockerImages and now.
+     */
+  } else if (
+    key === (await restoreCache([""], key, [], { lookupOnly: true }))
```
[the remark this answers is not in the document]
mwarres: Alas, Prettier disagrees. For what it's worth, I like you suggested aesthetic better.

src/docker.test.ts
```diff
@@ -87,14 +87,15 @@ describe("Docker images", (): void => {
   const assertSaveDockerImages = (
```
Kurt-von-Laven, line 87: Commit message: "save the cache" --> "attempt to save the cache" and "unnecessary cache saves" --> "unnecessary cache save attempts" since GitHub Actions doesn't allow you to overwrite a cache key without deleting the previous cache first.

src/docker.test.ts
```diff
@@ -117,14 +119,19 @@ describe("Docker images", (): void => {
     if (!cacheHit) {
       core.getInput.mockReturnValueOnce(readOnly.toString());
       if (!readOnly) {
-        core.getState.mockReturnValueOnce(preexistingImages.join("\n"));
-        const images = preexistingImages.concat(newImages);
-        util.execBashCommand.mockResolvedValueOnce(images.join("\n"));
+        if (prevSave) {
+          cache.restoreCache.mockResolvedValueOnce(key);
+        } else {
+          cache.restoreCache.mockResolvedValueOnce(key + "make keys mismatch");
```
Kurt-von-Laven, line 125: This is not a very realistic case to consider since `restoreCache` returns `undefined` when the cache isn't present.

Nit (moot): prefer a more natural-looking string that has a space after `key`. You may find it easier to remember this if you use string interpolation rather than string concatenation.

src/docker.test.ts
```diff
@@ -230,24 +247,29 @@ describe("Docker images", (): void => {
   );
 
   testProp(
-    "are saved unless cache hit, in read-only mode, or new Docker image list is empty",
+    "are saved unless cache hit, in read-only mode, a previous save occurs on " +
```
Kurt-von-Laven, line 250: I would summarize the third case as "cache already saved" and keep the comma before "or" for consistency.

src/docker.test.ts
```diff
@@ -264,10 +288,14 @@ describe("Docker images", (): void => {
...
       examples: [
-        ["my-key", false, false, [["preexisting-image"], ["new-image"]]],
-        ["my-key", false, false, [["preexisting-image"], []]],
-        ["my-key", false, true, [["preexisting-image"], ["new-image"]]],
-        ["my-key", true, false, [["preexisting-image"], ["new-image"]]],
+        ["my-key", false, false, false, [["preexisting-image"], ["new-image"]]],
+        ["my-key", false, false, false, [["preexisting-image"], []]],
+        ["my-key", false, true, false, [["preexisting-image"], ["new-image"]]],
+        ["my-key", true, false, false, [["preexisting-image"], ["new-image"]]],
+        ["my-key", false, false, true, [["preexisting-image"], ["new-image"]]],
```
Kurt-von-Laven, line 295: This is the only new example that is needed to ensure test coverage. The examples aren't used to cover all cases since fast-check is really good at that. They are just used to ensure that every line of production code gets hit.

src/docker.test.ts
```diff
@@ -87,14 +87,15 @@ describe("Docker images", (): void => {
...
     cacheHit: boolean,
     readOnly = false,
+    prevSave = false,
   ): void => {
     expect(core.getInput).nthCalledWith<[string, InputOptions]>(1, "key", {
       required: true,
     });
     expect(core.getState).nthCalledWith<[string]>(1, docker.CACHE_HIT);
     if (!cacheHit) {
       expect(core.getInput).lastCalledWith("read-only");
```
Kurt-von-Laven, line 97: We also want to assert that `cache.restoreCache` was called with the expected arguments if `!readOnly`.

Kurt-von-Laven (changes requested):

src/docker.ts
```diff
@@ -36,6 +36,18 @@ const saveDockerImages = async (): Promise<void> => {
...
+  ) {
+    info(
+      "A cache miss occurred during the initial attempt to load Docker " +
+        `images. Subsequently a cache with a matching key, ${key}, was saved. ` +
+        "This can occur when docker-cache is used by multiple jobs run in " +
+        "parallel. Not saving cache.",
```
Kurt-von-Laven, line 49: Nit: I have a tendency to limit self-references to the name of a tool because it makes it more tedious to change in the future (e.g., "This can occur when run in parallel.". That doesn't seem very likely to me here, but it's also a way to make this log statement a little less verbose. It might also help to join the first two sentences via a "but" to clarify their logical relationship.

src/docker.test.ts
```diff
@@ -86,7 +86,9 @@ describe("Docker images", (): void => {
```
Kurt-von-Laven, line 86: Commit message (nit): "was being run in parallel for multiple jobs" --> "was run in parallel" for brevity

Kurt-von-Laven (changes requested):

Kurt-von-Laven (approved):
**LGTM**

PR #516, opened 2023-04-21
style(util): Prefer `export type` to `export`

Kurt-von-Laven:
Clarify and enforce that `ConsoleOutput` is a type rather than a value.


mwarres (approved):
LGTM!

PR #526, opened 2023-04-27
ci(MegaLinter): Don't suppress pre-command errors

Kurt-von-Laven:
If a pre-command fails, fail MegaLinter, and display an error message rather than swallow the error.


mwarres (approved):
LGTM!

PR #528, opened 2023-04-27
ci(tsconfig): Include `types` directory and Jest config

Kurt-von-Laven:
Include the `types` directory and Jest config in the TypeScript build to ensure that they compile.


mwarres (approved):
LGTM!

PR #530, opened 2023-04-27
style(util): Prefer `interface` to `type`

Kurt-von-Laven:
The two are often interchangeable, but prefer interfaces when either will do since they are defined statically rather than imperatively, making them slightly simpler.


mwarres (approved):
LGTM!

PR #531, opened 2023-04-28
test(docker): Key mocks by reference, not name

Kurt-von-Laven:
In Jest, mock names make poor keys for identifying mock instances since they are not guaranteed to be unique and default to `"jest.fn()"` unless explicitly specified. All of our mocks are hence currently named `"jest.fn()"`. The key collisions resulted in all call orders after the first being `undefined`, causing `assertCalledInOrder` to always succeed vacuously. Key mocks by reference instead so `assertCalledInOrder` asserts that mocks are called in the specified order as intended.


mwarres (approved):
LGTM!

PR #533, opened 2023-04-28
test(docker): Check if mock called too few times

Kurt-von-Laven:
Fail fast since the call order is `undefined` in this case, eliminating a dangerous cast from `number | undefined` to `number`.

### DIFF
--- a/src/docker.test.ts
+++ b/src/docker.test.ts
@@ -21,16 +21,23 @@ const core = jest.mocked(await import("@actions/core"));
 const util = jest.mocked(await import("./util.js"));
 const docker = await import("./docker.js");
 
+const fail = (message: string): never => {
+  throw new Error(message);
+};
+
 // Expect the given mocks were called in the given order.
 const assertCalledInOrder = <T extends FunctionLike>(
   ...mocks: jest.MockedFunction<T>[]
 ): void => {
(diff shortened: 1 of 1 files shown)


mwarres (commented):
I'm not getting the advantage of introducing the variable `index` and using it within the error message.  To me, the idea of using reference to the mock function (`jest.MockedFunction<T>`), which I believe would be the name of the original function we're testing, seems like it would provide more helpful error messages.  Can you explain your rationale?

Kurt-von-Laven:
References aren't strings. The `toString()` method on a `jest.MockedFunction<T>` returns the source of the Jest code that created it: `"function getInput() {  return mockConstructor.apply(this,arguments);}"`. This behavior is [inherited from `Function`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString#description):

> For user-defined `Function` objects, the `toString` method returns a string containing the source text segment which was used to define the function.

The `getName()` method on a `jest.MockedFunction<T>` returns `"jest.fn()"` unless the name is explicitly set, which adds a lot of boilerplate.

mwarres:
Oh bummer!! Ok, index it is! Thanks for explaining!

mwarres (approved):
LGTM!

PR #534, opened 2023-04-28
style(util): Prefer `as` to angle brackets

Kurt-von-Laven:
Angle bracket casts can create confusion with generics.


mwarres (approved):
LGTM!

PR #535, opened 2023-04-29
test(integration): Prefer `??` to `||`

Kurt-von-Laven:
Prefer the nullish coalescing operator, `??`, to the logical OR operator, `||`, because the former only coalesces on `null | undefined` rather than on any falsy value.


mwarres (approved):
LGTM!

PR #536, opened 2023-04-29
ci(ESLint): Enable type-checking and strict rules

Kurt-von-Laven:
Automatically detect as many issues as possible.


mwarres (approved):
LGTM!

PR #539, opened 2023-04-29
test(integration): Prefer `Map` to `Object`

Kurt-von-Laven:
`Object` has a prototype that includes default keys that can potentially lead to collisions when used as a map.


mwarres (approved):
LGTM!

PR #540, opened 2023-04-29
test(integration): Rename `key` to `primaryKey`

Kurt-von-Laven:
For consistency, refer to the key parameter to the `setKey` helper function by the same name used for the corresponding parameter to the `cache.saveCache` and `cache.restoreCache` mocks.


mwarres (approved):
LGTM!

PR #541, opened 2023-05-02
test(docker): Prefer `concat` to `...`

Kurt-von-Laven:
The former has better performance than the spread operator for concatenating arrays.


mwarres (approved):
LGTM!

PR #549, opened 2023-05-04
test(docker): Eliminate unnecessary deduplication

Kurt-von-Laven:
`preexistingImages` and `newImages` are already guaranteed not to contain any duplicates either within or between them.


mwarres:
Ohhhhhhhh! Got it! Thanks!

mwarres (approved):
LGTM!

PR #550, opened 2023-05-04
test(action): Prefer `fullUnicodeString` to `string`

Kurt-von-Laven:
The `string` fast-check arbitrary only samples printable ASCII characters. Thus, it may overlook bugs that are only exposed by Unicode characters.


mwarres (approved):
LGTM!

PR #551, opened 2023-05-04
test(docker): Refine Docker images arbitraries

Kurt-von-Laven:
Introduce `uniquePair` arbitrary in `src/arbitraries/util.ts` that generates a 2-tuple of arrays with no overlap. Docker images can either be preexisting, that is present during the main step, or new, that is only present during the post step. Hence, restrict fast-check to cases without duplicates between preexisting and new Docker images. Previously, these duplicates were filtered out after the fact. Move this filtering into the arbitrary to produce more intuitive counterexamples. Modify the example that tested handling of duplicates accordingly.

Introduce `dockerImages` arbitrary in `src/arbitraries/util.ts` that forbids the empty string and newlines. Docker image names can neither be empty nor contain newlines, so there's no value in testing these cases that complicate test logic. Remove the no-longer-needed helper function `joinAndSplit`, which previously accounted for these edge cases.

### DIFF
--- a/src/arbitraries/util.ts
+++ b/src/arbitraries/util.ts
@@ -1,4 +1,12 @@
-import { constantFrom, fullUnicodeString, record } from "fast-check";
+import {
+  constantFrom,
+  fullUnicode,
+  fullUnicodeString,
+  record,
+  stringOf,
+  tuple,
+  uniqueArray,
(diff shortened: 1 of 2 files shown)


mwarres (approved):
LGTM! Nice change!

PR #575, opened 2023-07-21
Add possibility to specify images to cache

valentyn-red-bird:


action.yaml
```diff
@@ -11,6 +11,12 @@ inputs:
       are not supported, because partial cache restoration leads to a "snowball"
       effect.
     required: true
+  included-images:
+    description: >
+      If "all", all images will be included.
+      Otherwise, specify precisely all needed images to be included separated with spaces.
+    required: true
```
Kurt-von-Laven, line 18: I may be missing something, but I think this should be `false` since, when not specified, the default behavior ought to be the behavior that predates this PR.

src/docker.ts
```diff
@@ -29,6 +29,8 @@ const loadDockerImages = async (): Promise<void> => {
 
 const saveDockerImages = async (): Promise<void> => {
   const key = getInput("key", { required: true });
+  const includedImages = getInput("included-images", { required: true });
```
Kurt-von-Laven, line 32: I don't see the rationale for specifying this argument as required.

src/docker.ts
```diff
@@ -42,7 +44,9 @@ const saveDockerImages = async (): Promise<void> => {
     const images = await execBashCommand(LIST_COMMAND);
     const imagesList = images.split("\n");
     const newImages = imagesList.filter(
-      (image: string): boolean => !preexistingImages.includes(image)
+      (image: string): boolean =>
+        !preexistingImages.includes(image) &&
+        includedImagesList.includes(image)
```
Kurt-von-Laven, line 49: To pass CI/CR, you will need test coverage for the (currently unhandled) case where the default value is passed as well as the case where it is overridden by the user. I think users will find it counter-intuitive if pre-existing images are still filtered out when they specified explicitly what to cache.

src/docker.ts
```diff
@@ -52,6 +56,8 @@ const saveDockerImages = async (): Promise<void> => {
           "will be saved."
       );
       const newImagesArgs = newImages.join(" ");
+      info("newImagesArgs");
+      info(newImagesArgs);
```
Kurt-von-Laven, line 60: Was this left over from debugging? `execBashCommand` already logs the command passed to it, which contains this string.

action.yaml
```diff
@@ -11,6 +11,12 @@ inputs:
       are not supported, because partial cache restoration leads to a "snowball"
       effect.
     required: true
+  included-images:
+    description: >
+      If "all", all images will be included.
```
Kurt-von-Laven, line 16: I would change the default to `"new"` and clarify that we, to quote the README: "filter out Docker images that are present before the action is run, notably those [pre-cached by GitHub actions](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cached-docker-images); only save Docker images pulled or built in the same job after the action is run." I don't see much reason to support `"all"` unless we get some use case for that from folks using self-hosted runners that aren't pre-caching Docker images.

action.yaml
```diff
@@ -11,6 +11,12 @@ inputs:
       are not supported, because partial cache restoration leads to a "snowball"
       effect.
     required: true
+  included-images:
```
Kurt-von-Laven, line 14: I like the proposed name. Just gotta document it in the README too.

action.yaml
```diff
@@ -11,6 +11,12 @@ inputs:
       are not supported, because partial cache restoration leads to a "snowball"
```
Kurt-von-Laven, line 11: Nit (commit message): the scope should be "action" rather than "ci" since "ci" within this repository would refer to the CI pipeline for this action itself. (In fairness, this action is run as part of its own CI... he he.)

Kurt-von-Laven (changes requested):
This is a good concept and has the backing of multiple previous independent feature requests. Admittedly those were already addressed by filtering out pre-existing images, but in any case I am sold on this feature being worthy of inclusion at this point. Let me know if you need any help addressing the issues detected by CI after taking a look at our [contributing guide](https://github.com/ScribeMD/.github/blob/main/CONTRIBUTING.md). You will be able to address the security vulnerabilities simply by rebasing.

kivra-pauoli:
@Kurt-von-Laven, this pull request seems to have become stale.

I'd be willing to recreate it and push for approval. My use case is similar, but:

1. I have a very large number of images
2. they're named in a way that I could filter by regex (e.g. prefix)

I could implement regex based filtering on top of what is, too (?)

Thanks for the action, I can't wait to have it work for my use case 😄

PR #581, opened 2023-07-29
chore(Renovate): Add mailmap for Mend Renovate

Kurt-von-Laven:
The Renovate Bot Git user was renamed to Mend Renovate, so apply the rename to prior commits using Git mailmap.


mwarres (approved):
LGTM!

PR #602, opened 2023-08-04
chore(@types/prettier): Remove no longer needed dependency

Kurt-von-Laven:
Prettier 3+ includes its own type definitions, so `@types/prettier` is no longer needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "@types/eslint": "8.44.1",
     "@types/jest": "29.5.3",
     "@types/node": "18.17.2",
-    "@types/prettier": "2.7.2",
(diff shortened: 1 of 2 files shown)


mwarres (commented):
The commit looks good. However, the pull request title says "Remove no longer dependency", which doesn't make sense to me. I'm thinking you meant "no longer needed" but ran into character count issues? The way you dealt with it in the commit title, i.e. "Remove unneeded dependency" makes sense to me. Recommend changing it to that unless you have a rationale for this.

Kurt-von-Laven:
Good catch; fixed the typo in the PR title!

mwarres (approved):
LGTM!

PR #610, opened 2023-08-09
chore(Poetry): Group dev deps with updated syntax

mwarres:
Previously, development dependencies were grouped under `[tool.poetry.dev-dependencies]` in `pyproject.toml`. This syntax will soon be deprecated. Group development dependencies under `[tool.poetry.group.dev.dependencies]`, the updated syntax as of Poetry v1.2.0.


Kurt-von-Laven (changes requested):
Commit message: the [official Poetry docs](https://python-poetry.org/docs/managing-dependencies/#dependency-groups) say that the old syntax will be deprecated soon (but is not currently). Worth noting that the new syntax was released in Poetry 1.2.0.

mwarres:
@Kurt-von-Laven Are you able to take a look at this and see if it's ok?

Kurt-von-Laven (approved):
**LGTM**

poetry.lock
```diff
@@ -183,13 +183,13 @@ testing = ["covdefaults (>=2.3)", "coverage (>=7.3)", "diff-cover (>=7.7)", "pyt
 
 [[package]]
 name = "identify"
-version = "2.5.27"
+version = "2.5.28"
```
Kurt-von-Laven, line 186: Did you run `poetry update` locally? Please remove unrelated changes from this pull request.

Kurt-von-Laven (changes requested):

poetry.lock
mwarres: Just `poetry lock` and `poetry install`. I'll remove the change though. Thanks for catching this.

Kurt-von-Laven: `poetry lock` runs `poetry update`, which is why [our pre-commit hook instead runs `poetry lock --no-update`](https://github.com/ScribeMD/pre-commit-hooks#poetry-lock) to prevent it from constantly performing lock file maintenance.

Kurt-von-Laven: Still the same problem as before.

mwarres: Oh darn. Sorry, just seeing this.

Kurt-von-Laven: Same problem again.

Kurt-von-Laven (approved):
**LGTM**

PR #642, opened 2023-09-04
chore(Vim): Alphabetize Yarn SDK for Vim

Kurt-von-Laven:
Facilitate maintenance by maintaining consistency between repositories.


mwarres (approved):
LGTM!

PR #643, opened 2023-09-04
docs(Yarn): Document need for `darwin` architecture

Kurt-von-Laven:
The Yarn licenses plugin requires that all dependencies are installed.


mwarres (approved):
LGTM!

PR #663, opened 2023-09-11
chore(VSCode): Remove `eslint.packageManager`

Kurt-von-Laven:
This setting is deprecated since the ESLint VSCode extension now detects the package manager automatically.


mwarres (approved):
LGTM!

PR #664, opened 2023-09-11
test(jest.config): Treat `*.ts` files as ESM

Kurt-von-Laven:
Jest began recommending use of this setting in TypeScript ESM projects when it was introduced in v27.


mwarres (approved):
LGTM

PR #671, opened 2023-09-13
chore(poetry.lock): Lock added MarkupSafe wheels

mwarres:
MarkupSafe, a transitive dependency of Commitizen, recently retroactively added some wheels to the already released MarkupSafe 2.1.3. Add the artifacts to our Poetry lock file to prevent the poetry-lock pre-commit hook from failing in CI.


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
```
Kurt-von-Laven, line 1: This change does the opposite of what it claims. It removes the wheels rather than adding them. The wheels were added in #658.

Kurt-von-Laven (changes requested):

PR #672, opened 2023-09-14
ci(python-safety-dependencies-check): Remove hook

Kurt-von-Laven:
Trivy and Grype were added to MegaLinter in v6.0.0 and v7.2.0, respectively. Both scan for security vulnerabilities in Poetry dependencies. Hence, we no longer require a dedicated pre-commit hook for this purpose.


mwarres (approved):
LGTM

PR #673, opened 2023-09-14
test(Jest): Add `jest:esm` script

Kurt-von-Laven:
The `test` script runs tsc, then Jest. Decompose the `test` script for readability and to facilitate only running Jest via `yarn run jest:esm`.


mwarres (approved):
LGTM

PR #674, opened 2023-09-14
test(Jest): Eliminate dependency on ts-node

Kurt-von-Laven:
Jest relies on ts-node to read its TypeScript config file directly. We already compile the file with tsc though, so eliminate the dependency, and read from the JavaScript config instead.


mwarres (approved):
LGTM!

PR #693, opened 2023-09-29
ci(workflow): Hasten checkout with treeless clones

Kurt-von-Laven:
In v4.1.0, actions/checkout recently introduced support for Git's partial clones. Partial clones are smaller than full clones (`fetch-depth: 0`), because they don't clone historical blobs and/or trees. Partial clones do clone all historical commits and tags though, so they are larger than depth 1 shallow clones, the default when `fetch-depth` isn't specified. In partial clones, unlike shallow clones, Git operations will typically fetch data that isn't available locally as needed. Hence, prefer treeless clones, the smallest partial clones, to full clones, but continue using shallow clones when historical commits and tags aren't needed (e.g., by Commitizen).


Kurt-von-Laven:
I experimented with both treeless and blobless clones, and they increased the runtime of the entire workflow by about 50% and 10%, respectively, so I am closing this pull request.

PR #721, opened 2023-11-10
chore(@types/node): Bump from v18.18.5 to v20.9.0

Kurt-von-Laven:
We are already using Node.js 20.9.0 via asdf, so use the corresponding version of `@types/node`.


mwarres (approved):
LGTM

PR #730, opened 2023-11-20
chore(Yarn): Stop tracking `json5@2.2.3` in cache

Kurt-von-Laven:
We instruct Git to ignore the Yarn cache since we aren't using Yarn Zero-installs, so don't track anything in `.yarn/cache/`. These files are automatically populated locally on `yarn install`.


mwarres (approved):
LGTM

PR #766, opened 2023-12-19
ci(markdownlint): Replace deprecated rule alias

Kurt-von-Laven:
The `-header` aliases were deprecated in markdownlint v0.9.0; `-heading` is recommended instead.


mwarres (approved):
LGTM

PR #768, opened 2023-12-19
fix(Node.js): Upgrade from node16 to node20

### LINKED ISSUE #613: Node version mismatch
There is a mismatch between the node version specified in `package.json` and the actual node version when used as a part of GitHub workflow:

https://github.com/ScribeMD/docker-cache/blob/8cd7495f37b9e168dc6461e56ca302c4e8f6ddfd/package.json#L17-L20

https://github.com/ScribeMD/docker-cache/blob/8cd7495f37b9e168dc6461e56ca302c4e8f6ddfd/action.yaml#L23-L27

Kurt-von-Laven:
Closes #613.

We are already using Node.js 20.10.0 via asdf, so use the same major version of Node.js in the action.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -21,7 +21,7 @@ outputs:
       The ferried output of the official GitHub cache action. True on cache hit
       (even if the subsequent docker load failed) and false on cache miss.
 runs:
-  using: node16
+  using: node20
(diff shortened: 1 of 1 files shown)


mwarres (approved):
LGTM

PR #769, opened 2023-12-20
ci(Trivy): Port `.trivyignore` to `.trivyignore.yaml`

Kurt-von-Laven:
Use the more expressive YAML config format recently introduced in v0.45.0.


mwarres (approved):
LGTM

PR #810, opened 2024-03-20
chore(tsconfig): Upgrade from `node18` to `node20`

Kurt-von-Laven:
We already upgraded from Node.js v18 to v20, so use the appropriate base TypeScript config. The tsconfigs presently happen to be identical, but that could change, and it's confusing to use `node18` while we are running Node.js v20.


mwarres (approved):
LGTM

PR #815, opened 2024-03-24
chore(Poetry): Disable package mode

Kurt-von-Laven:
We only use Poetry for dependency management, not packaging.


mwarres (approved):
LGTM

PR #825, opened 2024-07-17
feat(docker): Compress docker images archive

madeindra:
Github action limits the size of cache to 10GB at max. For some repositories that has a lot of docker images built/used, this could causes "out of space" error. This error not only cause the caching to fail, but also the whole pipeline to fail even though the other jobs are completed successfully.

To prevent above issue, the docker images archive could further be compressed using `gzip`. This aligns to the available options of [docker save](https://docs.docker.com/reference/cli/docker/image/save/#save-an-image-to-a-targz-file-using-gzip) and [docker load](https://docs.docker.com/reference/cli/docker/image/load/#load-images-from-stdin).


madeindra:
I realized this would be slower, thus defeats the purpose of docker images caching

